PR #5, opened 2016-03-23
Adding DI system

fidian:
Adding dizzy as a dependency.


AbsentSemicolon:
I agree that Dizzy is a good module to use for DI.

PR #7, opened 2016-03-23
Rudimentary health check and framework for a server

fidian:
This is a proof of concept to get a health check up and running for use by other software that will monitor this application.

The server portion is only partially implemented (enough to log that it isn't starting).

Mostly this commit is intended to be a code review and explanation of the direction we could (should?) take with the codebase.

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -1,0 +1,9 @@
+"use strict";
+
+var container;
+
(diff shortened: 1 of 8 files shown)


AbsentSemicolon:
Looks go to merge in.

PR #9, opened 2016-03-24
Dizzy was updated and this bug no longer exists.

fidian:
Removing the workaround and updating the minimum version required in package.json.


AbsentSemicolon:
Glad this was fixed.

PR #11, opened 2016-03-24
Adding web server and adjusting health check to use it

AbsentSemicolon:


fidian:
We need to make webServer into a new instance whenever needed, not a singleton.  However, you're on that task already and there's nothing wrong with this code.

PR #12, opened 2016-03-24
Adding optional proxy protocol to the server

fidian:
This forces all traffic to have proxy protocol or not, thus having a separate
health check enpoint that did not have proxy protocol enabled is not possible.
Sadly, this means we drop the health check functionality.


AbsentSemicolon:
Reviewed and verified.

PR #13, opened 2016-03-24
Defaulting the port via environment variable

fidian:
Cloud9 prefers you use the $PORT environment variable.  Altered the code so
that it detects when the PORT environment variable exists and will start the
server on that port instead of using the port number in the config.


AbsentSemicolon:
Reviewed and verified.

PR #15, opened 2016-03-25
Bugfix for Proxy Protocol - Adjusting how we wrap HTTP

fidian:
This works differently than I had previously understood.  With this adjustment the code works as expected.

Previously I thought this monkey patched `http`.  Now I understand that it doesn't alter the original object (the one from `require()` and instead gives us back a new one.


AbsentSemicolon:
Reviewed and verified.

PR #16, opened 2016-03-28
Updating Dizzy

fidian:
This bumps the version so the spread operator is not used in the library.  That, in turn, allows the code to run on a Cloud9 VM without using nvm to update Node.  It also allows the code to run in AWS through Elastic Beanstalk.


AbsentSemicolon:
Reviewed and verified.

PR #17, opened 2016-03-28
Flipping the library exports to export the class

fidian:
Libraries were exporting a function that required use of a DI system in order to create the objects.  This commit pulls the container-related code out of the libraries, making them more usable to a wider audience.  All of the container-specific code is now in dependencies.js and whatever uses dependencies.js (like bin/server.js).

### DIFF
--- a/lib/api-server.js
+++ b/lib/api-server.js
@@ -11,6 +11,4 @@ class ApiServer {
     }
 };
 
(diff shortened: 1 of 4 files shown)


AbsentSemicolon:
I like this, much easier to see what the modules are.

PR #20, opened 2016-03-28
Adding debug logger

AbsentSemicolon:
This will be turned on through environment variables


fidian:
I see no security holes.  Approved.

PR #21, opened 2016-03-28
Initial commit of some API example communications

fidian:
Also rename docs to doc (per our discussion)


AbsentSemicolon:
Good explanation.

PR #22, opened 2016-03-29
Changing from Express to Restify

fidian:
Good selection of modules are available and this is geared towards REST APIs, which is exactly what we want.


AbsentSemicolon:
Looks good.

PR #23, opened 2016-03-29
Helmet

fidian:
Adds security features from Helmet.

This needs to be applied after #22 includes restify.  Do not apply this the other way.


AbsentSemicolon:
Helmet looks to give us some nice extra security. I like how we pull in each piece instead of blindly adding what Helmet decides to include.

PR #24, opened 2016-03-29
Avoid a full screen of error messages from npm

fidian:
When the server runs and doesn't properly parse, such as during development, we do not want the errors to show up on our screen.  Can not set a property nor do something special to avoid this.  See npm/npm#6124 for more information.


AbsentSemicolon:
Very nice. Glad to not be seeing that error when things don't go as planned.

PR #26, opened 2016-03-31
Removing HTTPS settings and unused https property

fidian:
Also modifying the base URI so that the code performs less work and the URL is in the format that we expect.


AbsentSemicolon:
Looks good.

PR #27, opened 2016-04-01
Pretty printing and enabling the node environment

fidian:
Pretty printing is pretty obvious.

The environment change lets us use "require" and "module" without any fussing.  Cloud9 may not require it, but local installs of eslint may (mine does).


AbsentSemicolon:
Looks easier to read, nice to have the environment in there too without fussing with node specific code.

PR #28, opened 2016-04-01
Adding a logger.console method for going to screen

fidian:
This is for making sure things do not go to a log file but are instead sent to the screen only.


AbsentSemicolon:
Interesting way of writing to the console.

PR #29, opened 2016-04-01
Adding middleware profiler and significant speed improvements

fidian:
Eliminating function calls, `.forEach()` callbacks, etc.  Simplified the baseUrl stuff.


AbsentSemicolon:
Nice to get rid of dependency if we don't need it.

PR #30, opened 2016-04-05
Adding Base64 encode/decode functions

fidian:
This will be useful for encrypted data that we hand back to a client and possibly for encoding other things like tokens or session information.


AbsentSemicolon:
Looks good, nice to have some en/decoding in place.

PR #31, opened 2016-04-05
Adding Testing Suite and Tests

AbsentSemicolon:
Using Jasmine-Node for testing suite and istanbul to provide coverage which is fed into Codecov to show in readme. Also including dependencies banners and integration into Travis-CI


spec/logger.spec.js
```diff
@@ -0,0 +1,57 @@
+describe("logger", function () {
+    "use strict";
+
+    var log, logger;
+
+    logger = require("../lib/logger");
```
fidian: Should be in the beforeEach for it to be reliably reset.

spec/logger.spec.js
```diff
@@ -0,0 +1,57 @@
...
+        });
+
+        it("prints to log from info", function () {
+            log.info("something");
+            expect(console.log).toHaveBeenCalledWith("something");
+        });
+
+        it("prints to error from error", function () {
+            log.error("something");
+            expect(console.error).toHaveBeenCalledWith("ERROR: something");
```
fidian: Should expect that the logged string has "ERROR:" in front.

spec/logger.spec.js
```diff
@@ -0,0 +1,57 @@
...
+        });
+
+        it("prints to error from warn", function () {
+            log.warn("something");
+            expect(console.error).toHaveBeenCalledWith("WARN: something");
+        });
+
+        it("prints to log from console", function () {
+            log.console("something");
+            expect(console.log.apply.mostRecentCall.args[1]).toEqual(["something"]);
```
fidian: Doesn't it work if you `expect(console.log).toHaveBeenCalledWith("something")`?

spec/web-server.spec.js
```diff
@@ -0,0 +1,259 @@
+describe("web-server", function () {
+    "use strict";
+
+    var fs, logger, restify, restMiddleware, webServer, WebServer;
+
+    WebServer = require("../lib/web-server");
```
fidian: Should move to the beforeEach.

spec/web-server.spec.js
```diff
@@ -0,0 +1,259 @@
...
+        spyOn(fs, "readFileSync");
+        spyOn(logger, "console");
+        spyOn(logger, "debug");
+        spyOn(logger, "error");
+        spyOn(logger, "info");
+        spyOn(logger, "warn");
+        spyOn(restify, "createServer").andReturn({});
+    });
+
+    beforeEach(function () {
```
fidian: Can this get merged with the above?  If so, can you drop some of those extra variables so they are scoped to only the callback inside the beforeEach?

spec/web-server.spec.js
```diff
@@ -0,0 +1,259 @@
...
+                return {};
+            }
+        };
+
+        restMiddleware = function () {
+
+        };
+
+        spyOn(fs, "readFileSync");
+        spyOn(logger, "console");
```
fidian: We should have a fake logger module that does the spying automatically for you.  `require("./mocks/logger-mock")` sort of thing.

spec/web-server.spec.js
```diff
@@ -0,0 +1,259 @@
...
+        spyOn(restify, "createServer").andReturn({});
+    });
+
+    beforeEach(function () {
+        webServer = new WebServer(fs, logger, restify, restMiddleware);
+    });
+
+    describe("constructor and configure", function () {
+        it("should set up properties without config options then calls configure", function () {
+            expect(webServer.config).toEqual({
```
fidian: We should be testing the behavior, not the internal properties.  Mostly things like "did the instance get created" and "Is it set up for HTTPS automatically" or stuff.  We should not peek inside the webServer.config (as we discussed earlier today).

spec/web-server.spec.js
```diff
@@ -0,0 +1,259 @@
...
+                handleUncaughtExceptions: true,
+                handleUpgrades: false,
+                httpsServerOptions: null,
+                name: "OpenToken API Test",
+                proxyProtocol: true,
+                spdy: true,
+                version: "2"
+            });
+
+            expect(fs.readFileSync).toHaveBeenCalledWith("./cert.crt");
```
fidian: Very fragile.  Ideally we would not need to mock out fs.

spec/logger.spec.js
AbsentSemicolon: If I don't spy on console.log.apply I can do expect(console.log).toHaveBeenCalledWith("something"). I'll go with this route.

spec/logger.spec.js
AbsentSemicolon: Have to leave variables out of before each so I can use them in tests, but the require I moved into beforeEach.

fidian:
I have reviewed the commits and the end result looks just fine to me.  A great start for getting testing going.  Need to get confirmation that another person approves of my changes.

AbsentSemicolon:
The changes not done by me look good and give a good basis for tests.

fidian:
Squashing!

PR #33, opened 2016-04-06
Splitting profiling code to MiddlewareProfiler

fidian:
This made it much easier to test.

### DIFF
--- a/lib/middleware-profiler.js
+++ b/lib/middleware-profiler.js
@@ -1,0 +1,139 @@
+"use strict";
+
+/**
+ * Monkey patches a server to enable profiling of middleware.  Does not
+ * profile middleware that was already added to the server.
+ *
+ * It only tests middleware that executes successfully.  Anything that
+ * throws will be lost.
+ *
+ * @return {Function}
+ */
+class MiddlewareProfiler {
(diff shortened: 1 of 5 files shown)


lib/middleware-profiler.js
```diff
@@ -0,0 +1,139 @@
...
+    toString() {
+        var lines;
+
+        lines = [];
+        Object.keys(this.profiles).forEach((name) => {
+            lines.push([
+                name,
+                this.profiles[name].hits + " hits",
+                Math.round(this.profiles[name].elapsed * 1000) + " ms",
+                Math.round(1000 * this.profiles[name].elapsed / this.profiles[name].hits) + " avg"
```
AbsentSemicolon: Would it be good to make sure we don't divide by 0? Saw an example of NaN in the test.

fidian: I don't care about NaN.  This is pretty quick and dirty profiling, just enough to prove what middleware is causing slowdowns.

AbsentSemicolon: Sounds good since we need this to be as quick as can be.

AbsentSemicolon:
Looks good, nice to have it separate for easier profiling and testing.

PR #34, opened 2016-04-06
Adding tests and fixing a Buffer-related bug

fidian:
When you take a buffer and construct a new buffer from it, Node seems to
copy the data over verbatim without converting the encoding.  To make
the Base64 data appear as binary again we first must convert the
incoming buffer to a string and then convert the string back to a Buffer
object with "base64" encoding.


AbsentSemicolon:
Looks good. Glad you caught this early.

PR #35, opened 2016-04-06
Moving back to the original package

fidian:
Special thanks to @ozinc for merging ozinc/node-restify-links#1


AbsentSemicolon:
Wonderful, it's good to get the official package coming back in again.

PR #36, opened 2016-04-08
Use async processes, starting with certificates and key files

fidian:
These changes do several things, and they are fairly tightly related.

1.  This adds the `bluebird` module from npm for promises.  They're faster than ES6 promises and will work just fine.  `bluebird` also has extra utility helpers that make dealing with promises easier.
2.  To be independent from the Promise library that is chosen for the project, a "promise" module was created that exposed specific functionality from `bluebird` that can be injected into the code.
3.  Web-server moved file reading to `app()` so the `configure()` method doesn't need to return a promise.
4.  All other functions in web-server now require promises because they depend on `app()`.  This was easy in some places and slightly more complex in others.
5.  To read the certificate and key files, `fs` is wrapped by `promise.promisifyAll()` and that defers to `bluebird` to promisify things.  This makes a promised version of `fs`.
6.  Testing asynchronous promises is a terrible pain, so a `promise-mock` module was made to mimic the `promise` module (step 2) and it uses a library I previously wrote that creates synchronous promises.  I had attempted in vain to use https://github.com/charleshansen/mock-promises for a day but couldn't get the promises injected nor resolved properly in the tests.  At this point, only 1 test was broken, which speaks VOLUMES about how we're designing things and the usefulness of the synchronous promise library.
7.  The final amount of changes to the web-server test were extremely minimal and focused mostly around the new parameter we inject and changing `fs.readFileSync()` to `fs.readFileAsync()`.


lib/web-server.js
```diff
@@ -7,34 +7,51 @@ class WebServer {
...
+
+        promise = this.app();
+
         if (arguments.length > 1) {
-            this.logger.debug("Adding middleware for route: " + a);
-            this.app().use(a, b);  // path and callback
-        } else {
-            this.logger.debug("Adding middleware");
-            this.app().use(a);  // a = callback
+            return promise.then((server) => {
```
AbsentSemicolon: This should set this.serverPromise and return server like below.

AbsentSemicolon:
Reviewed and looks good.

PR #38, opened 2016-04-08
Selecting a different set of versions

fidian:
Test against first and last for each major release in recent history.
This should give us adequate coverage on the versions so we know what will and what won't break.


AbsentSemicolon:
Reviewed and looks good. Guess we didn't need to check so many version.

PR #39, opened 2016-04-11
Testing last functionality inside the lib directory

AbsentSemicolon:


spec/api-server.spec.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+
+describe("ApiServer", () => {
+    var apiServer, webServerMock;
+
+    beforeEach(() => {
+        var ApiServer, configMock;
+
+        ApiServer = require("../lib/api-server");
+        configMock = {
```
fidian: I would simplify this test by eliminating configMock from here entirely and moving server creation to within the test.  Then you can say things like `var config = {}` and `expect(webServerMock).toHaveBeenCalledWith(config);`.

spec/api-server.spec.js
```diff
@@ -0,0 +1,41 @@
...
+            "send"
+        ]);
+        next = jasmine.createSpy("nextMock");
+        expect(webServerMock.configure).toHaveBeenCalledWith({
+            "baseUrl": "https://localhost:8080/",
+            "port": 8443
+        });
+        expect(apiServer).toEqual(jasmine.any(Object));
+        expect(webServerMock.addRoute).toHaveBeenCalledWith("get", "/", jasmine.any(Function));
+        routeCall = webServerMock.addRoute.mostRecentCall.args[2];
```
fidian: The testing of the addition of the route should be in a separate test.  I would also recommend wrapping it in

    expect(() => {
        routeCall(req, res, next);
    }).not.toThrow();

spec/dependencies.spec.js
```diff
@@ -0,0 +1,21 @@
...
+
+describe("dependencies", () => {
+    var dependencies;
+
+    beforeEach(() => {
+        dependencies = require("../lib/dependencies");
+    });
+
+    it("resolves a dependency", () => {
+        expect(dependencies).toEqual(jasmine.any(Object));
```
fidian: This line doesn't make sense when you say you are testing if "it resolves a dependency".  I suggest moving it into its own test.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,74 @@
...
+                title: "self-discovery"
+            }
+        });
+    });
+
+    it("calls restMiddleware with https", () => {
+        configMock.https = true;
+        restMiddleware(configMock, ServerMock);
+
+        // Other calls are tested above
```
fidian: Since the setup is different, I am uncertain that it still does all of the normal things with HTTPS enabled.

You should pull out the tests to a function so you can say

    expectNormalMiddlewareWasCalled();

It could look like

    function expectNormalMiddlewareWasCalled() {
        expect(HelmetMock.frameguard).toHaveBeenCalled();
        ....
    }

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        ]);
+        req = {
+            href: function () {
+                return "/path"
+            }
+        }
+        spyOn(req, "href").andCallThrough();
+        next = jasmine.createSpy("nextMock");
+        restMiddleware(configMock, ServerMock);
+        expect(logger.debug).toHaveBeenCalled();
```
fidian: I really don't care if debug was called.  That's not part of the module's advertised functionality.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        expect(HelmetMock.hidePoweredBy).toHaveBeenCalled();
+        expect(HelmetMock.hsts).not.toHaveBeenCalled();
+        expect(HelmetMock.ieNoOpen).toHaveBeenCalled();
+        expect(HelmetMock.noCache).toHaveBeenCalled();
+        expect(HelmetMock.noSniff).toHaveBeenCalled();
+        expect(HelmetMock.xssFilter).toHaveBeenCalled();
+        expect(restifyLinks).toHaveBeenCalled();
+        standardLinks = ServerMock.use.mostRecentCall.args[0];
+        expect(standardLinks).toEqual(jasmine.any(Function));
+        standardLinks(req, res, next);
```
fidian: The testing of `standardLinks()` should be a separate test.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        expect(logger.debug).toHaveBeenCalled();
+        expect(HelmetMock.frameguard).toHaveBeenCalled();
+        expect(HelmetMock.ieNoOpen).toHaveBeenCalled();
+        expect(HelmetMock.hidePoweredBy).toHaveBeenCalled();
+        expect(HelmetMock.hsts).not.toHaveBeenCalled();
+        expect(HelmetMock.ieNoOpen).toHaveBeenCalled();
+        expect(HelmetMock.noCache).toHaveBeenCalled();
+        expect(HelmetMock.noSniff).toHaveBeenCalled();
+        expect(HelmetMock.xssFilter).toHaveBeenCalled();
+        expect(restifyLinks).toHaveBeenCalled();
```
fidian: Missing a line that looks like this

    expect(Helmet ock.hsts).not.toHaveBeenCalled();

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,74 @@
+"use strict";
+
+describe("restMiddleware", () => {
+    var configMock, HelmetMock, logger, restMiddleware, ServerMock, restifyLinks;
```
fidian: `HelmetMock` and `ServerMock` are not classes.

You didn't alphabetize here.

fidian:
I've made comments about the tests.  Would you please review them and get back to me?

spec/rest-middleware.spec.js
fidian: Nevermind - I missed this.

spec/api-server.spec.js
```diff
@@ -0,0 +1,46 @@
+"use strict";
+
+describe("ApiServer", () => {
+    var apiServer, ApiServer, config, webServerMock;
```
fidian: Only `ApiServer` and `webServerMock` should be here.  The rest are not touched in the `beforeEach()` and thus won't need to be in scope yet.

spec/api-server.spec.js
```diff
@@ -0,0 +1,46 @@
...
+        ]);
+        next = jasmine.createSpy("nextMock");
+
+        apiServer = new ApiServer({}, webServerMock);
+
+        expect(webServerMock.addRoute).toHaveBeenCalledWith("get", "/", jasmine.any(Function));
+        routeCall = webServerMock.addRoute.mostRecentCall.args[2];
+        expect(() => {
+            routeCall(req, res, next);
+        }).not.toThrow();
```
fidian: What is `routeCall()` supposed to do?  That bit should be tested here too.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,84 @@
...
+            "hsts",
+            "ieNoOpen",
+            "noCache",
+            "noSniff",
+            "xssFilter"
+        ]);
+        serverMock = jasmine.createSpyObj("serverMock", [
+            "use"
+        ]);
+        configMock = {
```
fidian: A mocked object isn't a value-only object.  You should not call this one a mock because there's no methods on it.

fidian: Additionally, it looks like you can just use `{}` or a hardcoded config instead of this config object in most places.  I would prefer that as it more distinctly calls out the bits that you need for the function to exhibit the desired behavior.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,84 @@
...
+        expect(helmetMock.hidePoweredBy).toHaveBeenCalled();
+        expect(helmetMock.ieNoOpen).toHaveBeenCalled();
+        expect(helmetMock.noCache).toHaveBeenCalled();
+        expect(helmetMock.noSniff).toHaveBeenCalled();
+        expect(helmetMock.xssFilter).toHaveBeenCalled();
+        expect(restifyLinks).toHaveBeenCalled();
+    }
+
+    it("calls restMiddleware without https", () => {
+        restMiddleware(configMock, serverMock);
```
fidian: To elaborate my previous point, I would change this line to read

    restMiddleware({}, serverMock);

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,84 @@
...
+        configMock = {
+            baseUrl: "http://localhost:8443",
+            https: false
+        };
+        restifyLinks = jasmine.createSpy();
+        logger = new LoggerMock();
+        restMiddleware = new RestMiddleware(helmetMock, logger, restifyLinks);
+    });
+
+    function expectNormalMiddlewareWasCalled () {
```
fidian: Please document this function, move it just below the `var` line to help conceptually simulate the function hoisting that JavaScript performs and then eliminate that one space before the open parenthesis.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,84 @@
...
+        expect(helmetMock.hsts).not.toHaveBeenCalled();
+    });
+    it("calls restMiddleware with https", () => {
+        configMock.https = true;
+        restMiddleware(configMock, serverMock);
+
+        expectNormalMiddlewareWasCalled();
+        expect(helmetMock.hsts).toHaveBeenCalled();
+    });
+    it("sets up link discovery", () => {
```
fidian: It doesn't set up "link discovery" (that might be used in a client but the server doesn't discover anything).  Instead this sets up self-discovery links.

spec/rest-middleware.spec.js
```diff
@@ -0,0 +1,84 @@
...
+        var next, req, res, standardLinks;
+        res = jasmine.createSpyObj("resMock", [
+            "links"
+        ]);
+        req = {
+            href: function () {
+                return "/path"
+            }
+        }
+        spyOn(req, "href").andCallThrough();
```
fidian: We want to leverage as much Jasmine as we can.

    req = jasmine.createSpyObj("req", [
        "href"
    ]);
    req.href.andReturn("/path");

fidian:
Looks good!  Squashing into a single commit.

PR #40, opened 2016-04-12
S3 Plugin for Opentoken.

AbsentSemicolon:


lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * @param {Object} awsSdk
+     * @param {Object} promise
+     */
+    constructor(awsSdk, promise) {
+        this.aws = awsSdk;
+        this.config = {};
+        this.configure({});
+        this.promise = promise;
+        this.s3Bucket = null;
```
fidian: This is not a bucket.

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+
+
+    /**
+     * Creates the S3 bucket we need for subsequent calls.
+     *
+     * @return {Object} s3Bucket
+     */
+    transit() {
+        if (! this.s3Bucket) {
+            this.s3Bucket = new this.aws.S3({params: {Bucket: this.config.bucket}});
```
fidian: Split the object literal onto multiple lines.

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+        this.s3Bucket = null;
+    }
+
+
+    /**
+     * Creates the S3 bucket we need for subsequent calls.
+     *
+     * @return {Object} s3Bucket
+     */
+    transit() {
```
fidian: Alphabetically, this should be much later in the file.

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Configures options for AWS.
+     *
+     * @param {Object} config
+     * @return {Object} this
+     */
+    configure(config) {
+        this.aws.config.region = config.region || "us-east-1";
+        this.config.bucket = config.bucket || null;
+        config = null;
```
fidian: Unnecessary line, unless you're intentionally freeing up some large amount of memory.

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+     * @return {Object} promise
+     */
+    fileDel(filename) {
+        var params;
+        
+        params = {
+            Key: filename
+        };
+        
+        return this.transit().deleteObjectAsync(params);
```
fidian: Could you embed the parameters in the call?  Saves on the variable, especially since there's no logic in here.

    return this.transit().deleteObjectAsync({
        Key: filename
    });

fidian: Same goes for other places in this file

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+        config = null;
+
+        return this;
+    }
+    
+    /**
+     * Deletes a single file from configured bucket.
+     *
+     * @param {string} filename
+     * @return {Object} promise
```
fidian:
```
@return {Promise} indicates success/error of operation
```

fidian: Please also make this change elsewhere in this file.

lib/storage/s3.js
```diff
@@ -0,0 +1,146 @@
...
+            Key: filename,
+            ServerSideEncryption: "AES256"
+        };
+
+        if (options && typeof options === "object") {
+            params.ContentType = options.contentType || params.ContentType;
+            params.Expires = options.expires || null;
+        }
+
+        return this.transit().putObjectAsync(params);
```
fidian: This is an example of where I would keep the `params` object.  You conditionally modify it in this method.

spec/api-server.spec.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+
+describe("ApiServer", () => {
+    var ApiServer, webServerMock;
+
+    beforeEach(() => {
+        ApiServer = require("../lib/api-server");
+        webServerMock = jasmine.createSpyObj("webServerMock", [
+            "configure",
+            "addRoute",
```
fidian: ... Am I just a stickler for alphabetization?

fidian: It is adding "spec/api-server.spec.js" though that's really not part of your commits.  This is an example of why I'd love to have you base work off of "master" in the future.

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
+"use strict";
```
fidian: Would you move this to `spec/storage/s3.spec.js` to mirror the layout in `lib/`?

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+
+describe("storage/s3", () => {
+    var AwsSdkMock, promiseMock, s3;
```
fidian: This is not a class.  You don't use "new aws", you use "new aws.S3".

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+     * Checks the result for the promise object passed in.
+     *
+     * @param {Object} results
+     * @param {boolean} successful
+     * @param {string} message
+     */
+    function checkPromiseResults(results, successful, message) {
+        expect(results.success).toBe(successful);
+        expect(results.value).toEqual(message);
+    }
```
fidian: Per our discussion ... here's the code you think might be better to use because it keeps the results in with the test in a more readily accessible way.

```
expect(s3.doTheThing().status()).toEqual({
    success: true,
    value: "successful file get"
});
```

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+            constructor(params) {
+                this.params = params;
+                this.deleteObjectAsync = jasmine.createSpy("deleteObjectAsync");
+                this.getObjectAsync = jasmine.createSpy("getObjectAsync");
+                this.listObjectsAsync = jasmine.createSpy("listObjectsAsync");
+                this.putObjectAsync = jasmine.createSpy("putObjectAsync");
+            }
+        }
+        AwsSdkMock = {
+            "S3": S3Fake,
```
fidian: Do you need to quote properties?

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+        };
+        s3 = new s3File(AwsSdkMock, promiseMock);
+    });
+    describe("configure", () => {
+        beforeEach(() => {
+            AwsSdkMock.S3 = jasmine.createSpy("s3Mock");
+        });
+        it("passes in configuration options for all", () => {
+            s3.transit();
+            expect(s3.aws.config.region).toBe("us-east-1");
```
fidian: I think that this `expect()` after the call to `.transit()` makes it look like `.transit()` is setting this property.  Perhaps this is more clear?

    expect(s3.aws.config.region).toBe("us-east-1");
    s3.transit();

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+            expect(AwsSdkMock.S3).toHaveBeenCalledWith({params: {
```
fidian: Spacing on the object literal isn't what I would like to see.

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+            }});
+        });
+        it("passes in configuration options for all", () => {
+            s3.configure({
+                region: "us-west-1",
+                bucket: "test-bucket"
+            });
+            s3.transit();
+            expect(s3.aws.config.region).toBe("us-west-1");
+            expect(AwsSdkMock.S3).toHaveBeenCalledWith({params: {
```
fidian: Object spacing again.  I could recommend a useful tool if it helps.  Could you go through the file and look for /. (slash followed by any character) and fix that for me?

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+                Bucket: "test-bucket"
+            }});
+        });
+    });
+    describe("fileDel", () => {
+        var transit;
```
fidian: Not necessary outside of the `beforeEach`

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+        
+        beforeEach(() => {
+            transit = s3.transit();
+            transit.getObjectAsync.andCallFake((params) => {
+                if (params.Key == "afile") {
+                    return promiseMock.resolve("got data");
+                }
+
+                return promiseMock.reject("error fetching file");
+            });
```
fidian: This is starting to get redundant.  You might want to just set up mocks in the initial `beforeEach()` for all of the tests.

```
class S3Fake {
    constructor(params) {
        this.params = params;
        [
            'deleteObjectAsync',
            'getObjectAsync',
            'listObjectAsync',
            'putObjectAsync'
        ].forEach((method) => {
            this[method] = jasmine.createSpy(method);
            this[method].andCallFake((params) => {
                return promiseMock.resolve(params);
            });
        }
    }
}
```

You don't test any failed promises, which is ok by me, so the fake methods attached just resolve the promise with the arguments.  With that in place you can check the arguments.

```
expect(s3.fileList().status()).toEqual({
    success: true,
    value: {
        Prefix: "accounts"
    }
});
```

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+        });
+        it("gets top level list", () => {
+            checkPromiseResults(s3.fileList(), true, "got list from top level");
+        });
+        it("passes in a prefix", () => {
+            checkPromiseResults(s3.fileList("accounts"), true, "got list from accounts");
+        });
+    });
+    describe("filePut", () => {
+        var transit, expectedOptions;
```
fidian: Both of these variables should not be defined here.  `transit` is only used in `beforeEach()` and `expectedOptions` is a bit weird to have here too.

`beforeEach()` should set up tests, but not perform testing.  You're setting up expectations in `beforeEach()` and that is a sign of a poorly designed test or an overly complicated method of testing some functionality.

spec/s3.spec.js
```diff
@@ -0,0 +1,192 @@
...
+        it("passes in contents as a buffer", () => {
+            var putReturn;
+            
+            putReturn = s3.filePut("string", new Buffer("this is a buffer", "binary"));
+            checkPromiseResults(putReturn, true, "puts file");
+        });
+        it("passes in options", () => {
+            var putReturn;
+
+            expectedOptions = {
```
fidian: Eeeeeek!  You're passing DATA INTO BEFOREEACH!

spec/storage/s3.spec.js
```diff
@@ -0,0 +1,173 @@
...
+            });
+        });
+    });
+    describe("get()", () => {
+        beforeEach(() => {
+            var transit;
+            
+            transit = s3.transit();
+            transit.getObjectAsync.andCallFake((params) => {
+                return promiseMock.resolve({Body: new Buffer("this is a buffer", "binary")});
```
fidian:
```
{
    Body: new Buffer("this has a newline after the open brace", "ascii")
}
```

spec/storage/s3.spec.js
```diff
@@ -0,0 +1,173 @@
...
+                success: true,
+                value: {
+                    Body: jasmine.any(Buffer),
+                    ContentType: "application/octet-stream",
+                    Key: "buffer",
+                    ServerSideEncryption: "AES256"
+                }
+            });
+        });
+        xit("passes in options", () => {
```
fidian: Disabled test!

fidian:
Looks great!  Thanks for going through the extensive, soul crushing review process.  Squashing this commit.

PR #41, opened 2016-04-15
Adding Open Token Date Class

AbsentSemicolon:
This also includes the use of Moment.js for time functions.


lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            this.moment = m.utc();
+        }
+
+
+        /**
+         * Checks if the set time is before the time passed in.
+         *
+         * @param {Date} otherDate
```
fidian: Should be passing an `OtDate` instance or possibly a `Date` object.

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+         * @return {boolean} whether the date is before
+         */
+        isBefore(otherDate) {
+            return this.moment.isBefore(otherDate);
+        }
+
+
+        /**
+         * Adds time to the set time.
+         * You can send in multiple parameters to move the date up.
```
fidian: "up" can be used in two different ways.

Closer:
> Oh crap!  They moved the project due date up on us again!

Farther:
> Let's move the project due date up beyond the freeze window.

Could you change the wording?

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+         *
+         *    {"days": 1, "hours": 5, "seconds": 3}
+         *
+         * @param {Object} spec
+         * @return {Moment}
```
fidian: I don't ever want to expose the `moment` object outside.  This should not return or `return this`

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+
+            b = new Buffer(4);
+            b.writeUInt32LE(this.moment.unix(), 0);
+
+            return b;
+        }
+
+
+        /**
+         * Gets the date from the set moment.
```
fidian: What format is this in?

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
+"use strict";
+
+module.exports = function (moment) {
+    /**
+     * Open Token Date Class to provide simple date functionality
+     * using Moment.js.
```
fidian: You should mention that this is UTC only.

spec/ot-date.spec.js
```diff
@@ -0,0 +1,99 @@
...
+            expect(result).toEqual(jasmine.any(String));
+            expect(result).toBe("a formatted date");
+        });
+    });
+    it("throws an error when calling class directly", () => {
+        expect(() => {
+            new OtDate();
+        }).toThrow();
+    });
+});
```
fidian: Need to test if the date is guaranteed to be in UTC.

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+            b.writeUInt32LE(this.moment.unix(), 0);
+
+            return b;
+        }
+
+
+        /**
+         * Gets the date from the set moment.
+         *
+         * @return {string}
```
fidian: Documentation is wrong.  This returns `{Date}`

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+         */
+        toDate() {
+            return this.moment.toDate();
+        }
+
+
+        /**
+         * Gets the set moment and returns a formatted string.
+         *
+         * @return {string} set time formatted in ISO 8601 format
```
fidian: What is ISO 8601 format?  You should have an example.

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+    };
+
+
+    /**
+     * Sets a time from a buffer on the OtDate class.
+     *
+     * @return {OtDate}
+     */
+    OtDate.fromBuffer = (b) => {
```
fidian: Sorry, I gave you this function and now I'm amending it.

```
OtDate.fromBuffer = (b, offset) => {
    return new OtDate(new moment().unix(b.readUInt32LE(offset || 0)));
};
```

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+        return new OtDate(new moment().unix(b.readUInt32LE(0)));
+    };
+
+
+    /**
+     * Sets a passed in time on the OtDate class.
```
fidian: Comment is wrong.

lib/ot-date.js
```diff
@@ -0,0 +1,111 @@
...
+         * @return {string} set time formatted in ISO 8601 format
+         */
+        toString() {
+            return this.moment.toString(this.moment.ISO_8601);
+        }
+    };
+
+
+    /**
+     * Sets a time from a buffer on the OtDate class.
```
fidian: Doesn't just set the time.  It sets the entire date.

spec/ot-date.spec.js
```diff
@@ -0,0 +1,99 @@
...
+                    });
+                });
+                this.isBefore = jasmine.createSpy("isBefore");
+                this.isBefore.andCallFake((date) => {
+                    return true;
+                });
+                this.toString = jasmine.createSpy("toString");
+                this.toString.andCallFake(() => {
+                    return "a formatted date";
+                });
```
fidian: Consider having the returned value as part of the list ...  Here is just a partial example

```
methods = {
    add: this,
    fromString: this,
    toDate: this,
    isBefore: true,
    toString: "a formatted date"
};
Object.keys(methods).forEach((method) => {
    this[method] = jasmine.createSpy(method).andReturn(methods[method]);
});

Comment (fidian) [inline spec/ot-date.spec.js]: I don't really see why you are mocking moment.

File: lib/dependencies.js [hunk 13]
```diff
@@ -14,6 +14,7 @@ container.instance("apiServer", require("./api-server"));
 container.register("base64", require("./base64.js"));
 container.singleton("logger", require("./logger"));
 container.instance("middlewareProfiler", require("./middleware-profiler"));
+container.register("otDate", require("./ot-date"));
```
fidian: The capitalization here and in the test doesn't match.  You should consider if you are testing something that you want to treat as a class (probably not) or something that you want to treat as an object (most likely) and then capitalize or uncapitalize to match.

spec/ot-date.spec.js
```diff
@@ -0,0 +1,99 @@
...
+            }
+        };
+        OtDate = require("../lib/ot-date")(MomentFake);
+    });
+    describe("fromDate()", () => {
+        it("gets an OtDate object back from a custom date/time", () => {
+            expect(OtDate.fromDate("2016-04-16")).toEqual(jasmine.any(OtDate));
```
fidian: There is no check to make sure the date value inside is accurate.

I was thinking about this and I think that the fromDate compliments toDate.  You might want to unmock moment and then organize tests like so:

```
it("goes to and from a date", () => {
    var d, expected;

    d = new Date();
    expected = d.toString();
    ot = OtDate.fromDate(d);
    expect(ot.toDate().toString()).toEqual(expected);
})
```

This particular example is especially bad because `.fromDate()` should operate on a `Date`, not a string.

spec/ot-date.spec.js
```diff
@@ -0,0 +1,99 @@
...
+        });
+    });
+    describe("now()", () => {
+        it("gets the OtDate object back", () => {
+            var anotherTime, currentTime, result;
+
+            currentTime = new Date();
+            result = OtDate.now();
+            anotherTime = new Date();
+            expect(result).toEqual(jasmine.any(OtDate));
```
fidian: This missed the point of `currentTime` and `anotherTime`.  You didn't validate that the date value of `result` is correct.

spec/ot-date.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+    describe("plus()", () => {
+        it("sets the current time to three hours ahead", () => {
+            var result;
+
+            result = OtDate.now().plus({
+                hours: 3
+            });
+
+            expect(result).toEqual(jasmine.any(Object));
```
fidian: This did not validate the value of the date.

fidian:
We worked on this together at the end so I'm good with these changes.

PR #43, opened 2016-04-20
Switching serialization to use OtDate

fidian:
Needed to also create an OtDateMock for simplicity.  Useful for
injecting into other places without worrying about its dependencies,
though it's not nearly as robust as the real OtDate.

### DIFF
--- a/spec/mock/ot-date-mock.js
+++ b/spec/mock/ot-date-mock.js
@@ -71,4 +71,4 @@ module.exports = {
     now: () => {
         return new OtDateMock(new Date());
     }
-}
+};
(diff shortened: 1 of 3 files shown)


spec/mock/ot-date-mock.js
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Creates an OtDateMock from the current time
+     *
+     * @return {OtDateMock}
+     */
+    now: () => {
+        return new OtDateMock(new Date());
+    }
+}
```
AbsentSemicolon: Missing a semicolon here.

fidian: Fixed with d6ff3d3

AbsentSemicolon:
Reviewed and looks good.

PR #44, opened 2016-04-20
Moving all async functions to *Async

fidian:
This is a cleanup task and I removed the test-promise class because it isn't needed any longer.


AbsentSemicolon:
Reviewed and looks good.

PR #46, opened 2016-04-25
Adding Account Create and Complete

AbsentSemicolon:
Made the files into a few commits instead of the numerous it was before.


config.json
```diff
@@ -1,4 +1,12 @@
 {
+    "account": {
+        "completeLifetime": {
+            "months": 6
+        },
+        "initiateLifetime": {
+            "hours": 1
+        }
+    },
```
fidian: The account.completeLifetime, account.initiateLifetime and htop settings should all be documented.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
+"use strict";
+
+/**
+ * What we initially send back to the client so they can start to initiate
+ * the complete procedure and save information.
+ *
+ * @typedef {accountService~initiateResult}
+ * @param {string} accountId
+ * @param {string} email
+ * @param {string} mfa
```
fidian: Should rename this to be more clear.   This is the MFA seed or key, not a MFA generated value.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+class AccountService {
+    /**
+     * Used for Dependency Injection and initial configuration
+     * of the storage plugin.
+     *
+     * @param {Object} config
+     * @param {Object} password
+     * @param {Object} storage
+     */
+    constructor(config, password, storage) {
```
fidian: Config is unused.

lib/account/account-manager.js
```diff
@@ -0,0 +1,101 @@
...
+     * the Account record.
+     *
+     * Password generation should follow this format:
+     * hash(hash(password) + salt)
+     *
+     * @param {AccountManager~accountInfo} accountInfo
+     * @return {Promise}
+     */
+    complete(accountInfo) {
+        return this.accountService.get(accountInfo.accountId).then((accountFile) => {
```
fidian: I would not want this method to pass the accountFile back into complete.  Can the accountService cache the accountFile results instead and use those for future requests?  Caching for only the lifetime of the request, not longer.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} directory
+     * @param {Object} accountInfo
+     * @param {Object} accountFile
+     * @param {Object} options
+     * @return {Promise.<boolean>}
+     */
+    complete(directory, accountInfo, accountFile, options) {
```
fidian: Per my earlier comment, I don't want the accountFile added back here, otherwise the manager might want to set properties itself and not bother calling the service.

fidian: This should have "Async" at the end of the function name.

lib/account/account-manager.js
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Initiates the account creation and puts a file on the server
+     * with a hashed accountId.
+     *
+     * TODO: Validate email before doing anything else
+     *
+     * @param {Object} accountInfo
+     * @return {Promise.<Object>} accountService~initiateResult
+     */
+    initiate(accountInfo) {
```
fidian: This should be named initiateAsync.

lib/account/account-manager.js
```diff
@@ -0,0 +1,101 @@
...
+     * data off to Account service class for updating/putting data into
+     * the Account record.
+     *
+     * Password generation should follow this format:
+     * hash(hash(password) + salt)
+     *
+     * @param {AccountManager~accountInfo} accountInfo
+     * @return {Promise}
+     */
+    complete(accountInfo) {
```
fidian: Because this returns a Promise, we should name this completeAsync.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+     * so we never see the plain text version of it. This helps to keep
+     * the system more secure
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} directory
+     * @param {Object} accountInfo
+     * @param {Object} accountFile
+     * @param {Object} options
+     * @return {Promise.<boolean>}
```
fidian: Not sure why we return a boolean.  It doesn't seem to make sense.  Maybe resolve in the updated accountFile contents?

lib/account/account-manager.js
fidian: Also, with caching, we need to make sure we don't send back the object we are caching, otherwise things outside this code can update it.  I suggest caching the data before you parse the JSON.  Strings are not passed by reference.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Gets the account file from the directory.
+     *
+     * TODO: Decrypt data
+     *
+     * @param {string} accountId
+     * @return {Promise}
+     */
+    get(accountId) {
```
fidian: getAsync

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+
+    /**
+     * Gets the account file from the directory.
+     *
+     * TODO: Decrypt data
+     *
+     * @param {string} accountId
+     * @return {Promise}
```
fidian: This promise resolves in something, and that should be in this documentation.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+     * with a hashed accountId.
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} accountId
+     * @param {Object} accountInfo
+     * @param {Object} options
+     * @return {Promise.<Object>} accountService~initiateResult
+     */
+    initiate(accountId, accountInfo, options) {
```
fidian: initiateAsync.

lib/account/account-service.js
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Initiates the account creation and puts a file on the server
+     * with a hashed accountId.
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} accountId
+     * @param {Object} accountInfo
+     * @param {Object} options
+     * @return {Promise.<Object>} accountService~initiateResult
```
fidian: This line is wrong.  Here's the right one

```
@return {Promise.<accountService~initiateResult>}
```

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
+"use strict";
```
fidian: This file should be describing what it does.  If we require 10 factors for authentication, what factor is this one?

I'm thinking we want to rename the file from `hotp/two-factor-authenticator.js` to `mfa/hotp.js`.  Imagine we have all of these MFA solutions:

* Code sent via text message - `sms.js`
* Physical token mailed to a user - `hardware-token.js`
* Another person validating that the first one is trying to get in - `trusted-local-authority.js`
* Password (yeah, passwords are a factor in "multi factor") - `password.js`
* Voice call to their cell phone - `voice-call.js`
* and this module... - `virtual-token.js`?

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+
+    /**
+     * Generates the secret key the client will need to enter to
+     * initiate the two factor authentication.
+     *
+     * @return {Promise}
+     */
+    generateSecret() {
```
fidian: generateSecretAsync

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
...
+     * they can just scan the code.
+     *
+     * @param {string} secretKey
+     * @param {string} email
+     * @return {Promise.<string>}
+     */
+    generateQRCode(secretKey, email) {
+        email = email ? email : "";
+
+        return this.tfa.generateGoogleQRAsync("OpenToken IO", email, secretKey);
```
fidian: Does this just generate a URI?

I do NOT want any external service generating these MFA QR codes for us.

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
...
+     * Returns the image for seeing a QR code and can be then loaded into a
+     * browser or other application which can display the image.
+     * This makes it easier for the client to set up on their device as
+     * they can just scan the code.
+     *
+     * @param {string} secretKey
+     * @param {string} email
+     * @return {Promise.<string>}
+     */
+    generateQRCode(secretKey, email) {
```
fidian: generateQrCodeAsync

- needs Async
- QR when used in functions should be Qr.  Like "HTTP Server" becomes HttpServer.

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+
+    /**
+     * Verifies the token passed in is valid using the key
+     *
+     * @param {string} secretKey
+     * @param {string} toVerify token passed in from client
+     * @param {Object} options
+     * @return {Promise.<boolean>} whether the token validates
```
fidian: This doesn't look like it returns a promise to me and I really prefer to have this one not be async.

lib/account/account-manager.js
```diff
@@ -0,0 +1,101 @@
...
+                hotpOptions = this.config.hotp.previous;
+            }
+
+            if (! this.hotp.verifyToken(accountFile.mfa, accountInfo.previousMfa, hotpOptions)) {
+                throw new Error("Previous MFA Token did not validate");
+            }
+
+            if (! this.hotp.verifyToken(accountFile.mfa, accountInfo.currentMfa)) {
+                throw new Error("Current MFA Token did not validate");
+            }
```
fidian: With the default drift settings, we are able to enter the same code in both of these and it would still validate.  What we want to do is make the drift equal to the drift forward for these two and then loop backwards up to the allowed amount of drift backward.

lib/hotp/two-factor-authenticator.js
```diff
@@ -0,0 +1,85 @@
...
+            // the number of counters to check before what we're given
+            // default: 0
+            beforeDrift: options.beforeDrift || 0,
+            // and the number to check after
+            // default: 0
+            afterDrift: options.afterDrift || 0,
+            // if before and after drift aren't specified,
+            // before + after drift are set to drift / 2
+            // default: 0
+            drift: options.drift || 0,
```
fidian: Eliminate this option.  Do not allow three options to control drift.

lib/password.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+
+module.exports = function(crypto) {
+
```
fidian: No blank line here.  Space after "function" above.

lib/password.js
```diff
@@ -0,0 +1,24 @@
...
+     *
+     * @param {string} content
+     * @return {(string|boolean})
+     */
+    function hashContent(content) {
+        if (! content) {
+            return false;
+        }
+
+        return crypto.createHash("sha256").update(content).digest("base64");
```
fidian: Algorithm should come from config.

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,154 @@
...
+                this.verifyToken = jasmine.createSpy();
+                this.verifyToken.andCallFake((key, token, opts) => {
+                    if (token === "987654") {
+                        return false;
+                    }
+
+                    return true;
+                });
+            }
+        }
```
fidian: Pull these two out to separate mocks in spec/mocks/

Also, we are not using them as real classes.

hotpFake = jasmine.createSpyObj("HotpFake", [ ... ])

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,154 @@
...
+        class HotpFake {
+            constructor() {
+                this.generateSecret = jasmine.createSpy("generateSecret");
+                this.generateSecret.andCallFake(() => {
+                    return promiseMock.resolve("thisisasecrectcodefrommfa");
+                });
+                // This is a sync call
+                this.verifyToken = jasmine.createSpy();
+                this.verifyToken.andCallFake((key, token, opts) => {
```
fidian: Tip for the future - you can chain these two calls together.

this.verifyToken = jasmine.createSpy().andCallFake(...);

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,154 @@
...
+                expect(result).toEqual({
+                    accountId: jasmine.any(Buffer),
+                    email: "some.one@example.net",
+                    mfa: "thisisasecrectcodefrommfa",
+                    salt: jasmine.any(Buffer)
+                });
+                expect(result.accountId.length).toBe(24);
+                expect(result.salt.length).toBe(128);
+
+            }).then(done, done);
```
fidian: I don't understand how this is verifying that it called the service.  The manager's responsibility is only to apply business logic.  It only verifies that the tokens are right and sends an expires value to the service.  Nothing in here appears to be testing that functionality.  So you have the test coverage but appear to have missed the point of the function.

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,154 @@
...
+        var expectError, fail;
+
+        beforeEach(() => {
+            expectError = (done, contains) => {
+                // Generate a function that asserts the result is an error
+                // and contains some text in the message.
+                return (err) => {
+                    expect(err).toEqual(jasmine.any(Error));
+                    expect(err.toString()).toContain(contains);
+                    done();
```
fidian: 1) This should just be a function, perhaps at line 91.
2) This is the third time we do this stuff.  It's time to make custom matchers and functions.  That should be a new task - don't hold up this patch but it should get addressed next.

spec/account/account-service.spec.js
```diff
@@ -0,0 +1,88 @@
...
+                    );
+                });
+                this.putAsync = jasmine.createSpy("putAsync");
+                this.putAsync.andCallFake((params) => {
+                    return promiseMock.resolve(() => {
+                        return true;
+                    });
+                });
+            }
+        }
```
fidian: Same comment about the class here.  We don't really use a class.  If we did, making a mock that others could use would likely be a better scenario.

spec/account/account-service.spec.js
```diff
@@ -0,0 +1,88 @@
...
+        config = {
+            storage: {
+                bucket: "some-place-wonderful"
+            }
+        };
+        password = jasmine.createSpyObj("password", [
+            "hashContent"
+        ]);
+        password.hashContent.andCallFake(() => {
+            return "hashedContent";
```
fidian: I don't see where this value is ever tested in your tests.  I also don't see where you confirm the right password was sent in to this function.

spec/hotp/two-factor-authenticator.spec.js
```diff
@@ -0,0 +1,89 @@
...
+        });
+        tfaMock.generateGoogleQRAsync.andCallFake(() => {
+            return promiseMock.resolve(
+                "data:image/png;base64,iVBORw0KGgoA....5CYII="
+            );
+        });
+        hotp = new HOTP(tfaMock, promiseMock);
+    });
+    describe(".generateSecret()", () => {
+        it("returns a key at the proper length", (done) => {
```
fidian: The length was not validated.

```
expect(tfaMock.generateKeyAsync).toHaveBeenCalledWith(12345);
```

spec/password.spec.js
```diff
@@ -0,0 +1,25 @@
...
+            "createHash",
+            "digest",
+            "update"
+        ]);
+        cryptoFake.createHash = jasmine.createSpy("crypto.createHash").andReturn(cryptoFake);
+        cryptoFake.digest = jasmine.createSpy("crypto.digest").andReturn("hashedcontent");
+        cryptoFake.update = jasmine.createSpy("crypto.update").andReturn(cryptoFake);
+        password = require("../lib/password")(cryptoFake);
+    });
+    it("hashes a passed in value", () => {
```
fidian: I would greatly prefer that you test the hashed value against an expected value.  Real hashes.  That's the point of this library.

lib/hotp/two-factor-authenticator.js
AbsentSemicolon: This generates an image file which can be copied and pasted into the address bar of browser or I guess parsed to generate the actual image in another application. Below is an example of what is generated. The code does not reach out to any other service to generate, it is all generated in the library.

data:image/png;base64,iVBORw0KGgoA....5CYII=

lib/dependencies.js
```diff
@@ -40,5 +45,6 @@ container.provide("moment", require("moment"));
 container.provide("proxywrap", require("findhit-proxywrap"));
 container.provide("restify", require("restify"));
 container.provide("restifyLinks", require("restify-links"));
+container.provide("tfa", require("2fa"));
```
fidian: Why isn't this "2fa" instead of "tfa"?

spec/account/account-service.spec.js
AbsentSemicolon: On line 50 I validate the result of passing in the password from what comes back from the service now. From a comment above instead of returning true, we pass back the data of what was sent in.

```
expect(result).toEqual({
    email: "some.one@example.net",
    password: "somereallylonghashedpassword"
});
```

lib/account/account-service.js
AbsentSemicolon: Config is used on the line 29 to configure the storage plugin. Because configure returns the object, I thought it would be ok to just pass config into the configure method and have it chained on where I do this.storage = storage.

lib/account/account-manager.js
AbsentSemicolon: Nice idea, it does seem better to not have to pass accountFile back in. I have it cached now in the service.

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,154 @@
...
+            },
+            hotp: {
+                previous: {
+                    afterDrift: 1,
+                    beforeDrift: 1,
+                    drift: 2
+                }
+            }
+        };
+        accountManager = new AccountManager(new AccountServiceFake, config, new HotpFake, otDateMock, randomMock, promiseMock);
```
fidian: accountManager = new AccountManager(new AccountServiceFake(), config, new HotpFake(), otDateMock, randomMock, prom....

You should use () whenever you use new.

AbsentSemicolon: Converted to not be classes in the tests anymore.

lib/dependencies.js
AbsentSemicolon: This is because we can't the namespace of an object start with a number.  I renamed it to twofa instead to keep the sounding the same as the library.

spec/password.spec.js
AbsentSemicolon: Made to use Crypto instead of a mock of crypto and testing against hash passed in.

lib/hotp/two-factor-authenticator.js
AbsentSemicolon: Look in lib/mfa/hotp.js

doc/password.md
```diff
@@ -0,0 +1,8 @@
+Password
+=======
+
+The `password` option in `config.json` allows you to set which algorythm we use to hash a value. Primarily this is being used for `accountId` when stored so we aren't using the original `accountId`.
```
fidian: s/algorythm/algorithm/

doc/password.md
```diff
@@ -0,0 +1,8 @@
...
+
+    "password": {
+        "hashAlgo": "sha256"
+    }
```
fidian: Would you add a bit about being able to find available hashing methods in `lib/ciphers-and-hashes.js`?

doc/account.md
```diff
@@ -0,0 +1,35 @@
...
+        "completeLifetime": {
+            "months": 6
+        },
+        "initiateLifetime": {
+            "hours": 1
+        }
+    }
+
+
+Lifespan
```
fidian: I don't see "lifespan" in the object.  I also don't see a goo description of what the lifespan is.  I think I know what you're trying to get at, so I'd suggest adding more words in the paragraph on line 21.

doc/account.md
```diff
@@ -0,0 +1,35 @@
...
+--------
+
+Account creation and use goes through several phases.
+
+Initiate
```
fidian: This should be seen as "underneath" lifespan, so H3.  You do that with hashes

```
### Initiate
```

doc/account.md
```diff
@@ -0,0 +1,35 @@
...
+--------
+
+The first part of an account's lifespan is `initiate`. This sets up a few pieces of data needed to properly generate and track an account. We generate an Account ID, MFA Key, abd Password Salt. An Email address must be passed in. These are then passed back to the user to store on their own and will be used to generate their password.
+
+This will save the account using a hash of the their `accountId`.
```
fidian: That doesn't read well.  How about this?

This will save the account to a file named after the hash of their `accountId`.

lib/account/account-manager.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+
+    /**
+     * Validates the previous and current MFA values are valid and sends
+     * data off to Account service class for updating/putting data into
+     * the Account record.
+     *
+     * Password generation should follow this format:
+     * // insert password generation here
```
fidian: Not inserted.  :-(

lib/account/account-manager.js
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param {AccountManager~accountInfo} accountInfo
+     * @return {Promise}
```
fidian: What does this promise resolve with?

lib/account/account-manager.js
```diff
@@ -0,0 +1,97 @@
...
+
+
+    /**
+     * Initiates the account creation and puts a file on the server
+     * with a hashed accountId.
+     *
+     * TODO: Validate email before doing anything else
+     *
+     * @param {Object} accountInfo
+     * @return {Promise.<Object>} accountService~initiateResult
```
fidian:
```
@return {Promise.<accountService~initiateResult>}
```

lib/account/account-service.js
```diff
@@ -0,0 +1,101 @@
...
+     * client. We need the client to create the password using their tools
+     * so we never see the plain text version of it. This helps to keep
+     * the system more secure
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} directory
+     * @param {Object} accountInfo
+     * @param {Object} options
+     * @return {Promise.<boolean>}
```
fidian: Doesn't resolve in a boolean.  Instead it resolves in the accountFile data.

lib/account/account-service.js
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Used for Dependency Injection and initial configuration
+     * of the storage plugin.
+     *
+     * @param {Object} config
+     * @param {Object} password
+     * @param {Object} storage
+     */
+    constructor(config, password, storage) {
+        this.accountFile = null;
```
fidian: The service shouldn't have state.

lib/account/account-service.js
```diff
@@ -0,0 +1,101 @@
...
+     *
+     * @param {string} directory
+     * @param {Object} accountInfo
+     * @param {Object} options
+     * @return {Promise.<boolean>}
+     */
+    completeAsync(directory, accountInfo, options) {
+        this.accountFile.password = accountInfo.password;
+
+        return this.storage.putAsync(directory, JSON.stringify(this.accountFile), options).then(() => {
```
fidian: Do a "get" and remove the caching mechanism you have for now.  We should have a cache that works throughout the lifetime of a request and is cleared after each request.  That's harder to do, so let's take the performance penalty of two S3 gets.

lib/account/account-service.js
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Initiates the account creation and puts a file on the server
+     * with a hashed accountId.
+     *
+     * TODO: Encrypt data
+     *
+     * @param {string} accountId
+     * @param {Object} accountInfo
+     * @param {Object} options
+     * @return {Promise.accountService~initiateResult}
```
fidian:
```
@return {Promise.<accountService~initiateResult>}
```

Missing < and >

lib/mfa/hotp.js
```diff
@@ -0,0 +1,69 @@
...
+ * application we want to verify the client has access to.
+ *
+ */
+class HOTP {
+    /**
+     * We want to promisfy all methods we can because some of them
+     * take some time to generate and we don't want code going on
+     * till we have the data we need back.
+     *
+     * @param {Object} tfa
```
fidian: You noticed this one.  It should be "twofa"

lib/mfa/hotp.js
```diff
@@ -0,0 +1,69 @@
...
+    constructor(twofa, promise) {
+        this.twofa = promise.promisifyAll(twofa);
+    }
+
+
+    /**
+     * Generates the secret key the client will need to enter to
+     * initiate the two factor authentication.
+     *
+     * @return {Promise}
```
fidian: What does this promise resolve with?

Why is the key hardcoded to 128?

lib/mfa/hotp.js
```diff
@@ -0,0 +1,69 @@
...
+     * browser or other application which can display the image.
+     * This makes it easier for the client to set up on their device as
+     * they can just scan the code.
+     *
+     * @param {string} secretKey
+     * @param {string} email
+     * @return {Promise.<string>}
+     */
+    generateQrCodeAsync(secretKey, email) {
+        email = email ? email : "";
```
fidian: Shorthand:

```
email = email || "";
```

lib/mfa/hotp.js
```diff
@@ -0,0 +1,69 @@
...
+
+        return this.twofa.generateGoogleQRAsync("OpenToken IO", email, secretKey);
```
fidian: OpenToken IO should be in the config.

Our name is "OpenToken.io".

lib/mfa/hotp.js
```diff
@@ -0,0 +1,69 @@
...
+     * @param {string} secretKey
+     * @param {string} toVerify token passed in from client
+     * @param {number} beforeDrift the number of counters to check before what we're given
+     * @return {boolean} whether the token is valid
+     */
+    verifyToken(secretKey, toVerify, beforeDrift) {
+        var opts;
+
+        opts = {
+            beforeDrift: +(beforeDrift) || 0
```
fidian: Don't need parenthesis.

lib/password.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+module.exports = function (config, crypto) {
+    /**
+     * Hashes the content passed, but if no content was passed
+     * just return false as there's nothing to do.
```
fidian: What is the use case for passing this function no content?

I think I'd rather either not check or throw an Error.

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        AccountManager = require("../../lib/account/account-manager");
+        otDateMock = require("../mock/ot-date-mock");
+        promiseMock = require("../mock/promise-mock");
+        randomMock = require("../mock/random-mock");
+        accountServiceFake = jasmine.createSpyObj("accountServiceFake", [
+            "completeAsync",
+            "getAsync",
+            "initiateAsync"
+        ]);
+        accountServiceFake.completeAsync = jasmine.createSpy("complete").andCallFake(() => {
```
fidian: The spy is already created.  You can use

```
accountServiceFake.completeAsync.andCallFake(...)
```

fidian: Same goes for the ones below.

fidian: Interestingly, these three spies don't match the names in the method.  You're making the complete, get, and initiate spies here but they are all async methods.

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        });
+        hotpFake = jasmine.createSpyObj("hotpFake", [
+            "generateSecretAsync",
+            "verifyToken"
+        ]);
+        hotpFake.generateSecretAsync = jasmine.createSpy("generateSecretAsync").andCallFake(() => {
+            return promiseMock.resolve("thisisasecrectcodefrommfa");
+        });
+        hotpFake.verifyToken = jasmine.createSpy("verifyToken").andCallFake((key, token, opts) => {
+            if (token === "987654") {
```
fidian: Abbreviate:

```
return token === "987654"
```

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,149 @@
...
+                expect(result.accountId.length).toBe(24);
+                expect(result.salt.length).toBe(128);
+
+            }).then(done, done);
+        });
+    });
+    describe(".completeAsync()", () => {
+        var expectError, fail;
+
+        // TODO: Create custom matchers
```
fidian: You should change TODO into TODONE

fidian: TODO means "for a future commit, not part of this merge request", right?

spec/account/account-manager.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        function expectError (done, contains) {
+            // Generate a function that asserts the result is an error
+            // and contains some text in the message.
+            return (err) => {
+                expect(err).toEqual(jasmine.any(Error));
+                expect(err.toString()).toContain(contains);
+                done();
```
fidian: Custom matchers can not call `done` for you.  Maybe a helper function or maybe use a matcher and manually call done.  To be determined LATER, not now.

```
jasmine.expectError(done, contains);
jasmine.fail(done)
```

spec/account/account-service.spec.js
```diff
@@ -0,0 +1,88 @@
...
+        var AccountService, config, password, promiseMock, storageFake;
+
+        AccountService = require("../../lib/account/account-service");
+        promiseMock = require("../mock/promise-mock");
+        storageFake = jasmine.createSpyObj("storageFake", [
+            "configure",
+            "getAsync",
+            "putAsync"
+        ]);
+        storageFake.configure = jasmine.createSpy("storageFake.configure").andCallFake(() => {
```
fidian: Simplify this as well.  This happens multiple times in this file:

```
storageFake.configure.andCallFake(() => { ... }
```

spec/mfa/hotp.spec.js
```diff
@@ -0,0 +1,77 @@
+"use strict";
+
+describe("HOTP", () => {
+    var hotp, twofaMock, promiseMock;
+
+    beforeEach(() => {
+        var HOTP;
```
fidian: `Hotp` instead of `HOTP` if it is a class, `hotp` for an instance or a dumb object.

spec/mfa/hotp.spec.js
```diff
@@ -0,0 +1,77 @@
...
+            );
+        });
+        hotp = new HOTP(twofaMock, promiseMock);
+    });
+    describe(".generateSecretAsync()", () => {
+        it("returns a key at the proper length", (done) => {
+            hotp.generateSecretAsync().then((result) => {
+                expect(result).toBe("thisIsAReallyLongKeyForMFA");
+            }).then(done, done);
+            expect(twofaMock.generateKeyAsync).toHaveBeenCalledWith(128);
```
fidian: This expect should be either moved up one line or up three lines.

fidian: Likewise we should do that for the rest of the tests.

spec/mock/random-mock.js
```diff
@@ -0,0 +1,34 @@
+"use strict"
+
+var random;
+
+random = jasmine.createSpyObj("random", [
+    "bufferAsync",
+    "password"
+]);
+
+random.bufferAsync = jasmine.createSpy("random.bufferAsync").andCallFake((size) => {
```
fidian: As mentioned earlier, you don't need to create a spy here because `.createSpyObj()` does that for you.

spec/password.spec.js
```diff
@@ -0,0 +1,23 @@
...
+        crypto = require("crypto");
+        config = {
+            password: {
+                hashAlgo: "sha256"
+            }
+        };
+        password = require("../lib/password")(config, crypto);
+    });
+    it("hashes a passed in value", () => {
+        expect(password.hashContent("rRTcBER_EiFUsRa34Hj5Zpok")).not.toBe("rRTcBER_EiFUsRa34Hj5Zpok");
```
fidian: What should it be?  I figure that would be a more accurate test.  Perhaps the code always returns `null`, in which case the test would pass and you'd get coverage, but it would not be a meaningful test.

lib/account/account-manager.js
AbsentSemicolon: It was there, but when I mistakenly did my work on a different branch and tried to merge it, it got lost and it's not on the other branch.  But I'll get it back in.

spec/account/account-manager.spec.js
AbsentSemicolon: Yeah, that's part of a future feature that was brought up as part of the pull request, but we aren't doing it now.

spec/account/account-manager.spec.js
AbsentSemicolon: Should be `return token !== "987654"` as I want the opposite.  Just the way I've constructed the tests using the verify token method.

spec/account/account-manager.spec.js
AbsentSemicolon: This was a complication from when I was editing on the wrong branch. It's corrected now.

lib/mfa/hotp.js
AbsentSemicolon: Have this in config.json now.

fidian:
I see some areas of improvement, but let's get this in.

PR #47, opened 2016-04-26
Export only factories

fidian:
This rewrites the libraries so they always export factories.  Those factories may export classes or plain objects.  When a class had no internal state, it was converted to a plain object.

This also caught a few problems with the testing that were invisible previously, such as places where we reached into the instance and checked internal properties instead of checking that the right things were performed on the dependencies that were passed in.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -2,66 +2,62 @@
 
 "use strict";
 
-class Logger {
-    constructor(config) {
-        this.config = config;
-    }
+module.exports = (config) => {
+    return {
+        /**
+         * Displays a message to the console only.
+         *
+         * @param {*} ...data All parameters are sent to console.log.
+         */
+        console: function (x) {
+            var args;
 
-    /**
-     * Displays a message to the console only.
-     *
-     * @param {*} ...data All parameters are sent to console.log.
-     */
-    console() {
-        var args;
-
-        args = [].slice.call(arguments);
-        console.log.apply(console, args);
-    }
+            args = [].slice.call(arguments);
+            console.log.apply(console, args);
+        },
 
 
-    /**
-     * Logs debug statements to stderr only if
-     * debug has been turned on in environment variables
-     * or config.
-     *
-     * @param {string} text
-     */
-    debug(text) {
-        if (this.config.debug) {
-            console.error("DEBUG: " + text);
+        /**
+         * Logs debug statements to stderr only if
+         * debug has been turned on in environment variables
+         * or config.
+         *
+         * @param {string} text
+         */
+        debug: (text) => {
+            if (config.debug) {
+                console.error("DEBUG: " + text);
+            }
+        },
+
+
+        /**
+         * Logs errors to stderr
+         *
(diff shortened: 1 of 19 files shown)


lib/logger.js
```diff
@@ -2,66 +2,62 @@
...
-        if (this.config.debug) {
-            console.error("DEBUG: " + text);
+module.exports = (config) => {
+    return {
+        /**
+         * Displays a message to the console only.
+         *
+         * @param {*} ...data All parameters are sent to console.log.
+         */
+        console: function (x) {
```
AbsentSemicolon: Do we need x? You don't do anything with it. Might have been left over when explaining the situation.

lib/storage/s3.js
```diff
@@ -1,142 +1,142 @@
...
-    delAsync(filename) {
-        return this.transit().deleteObjectAsync({
-            Key: filename
-        });
-    }
+        /**
+         * Configures options for AWS.
+         *
+         * @param {Object} config
+         * @return {Object} this
```
AbsentSemicolon: Should this be something like `@return {S3} this` since it's returning the class?

lib/rest-middleware.js
```diff
@@ -1,11 +1,11 @@
-function restMiddleware(helmet, logger, restifyLinks) {
+module.exports = function restMiddleware(helmet, logger, restifyLinks) {
```
AbsentSemicolon: Do we need this to day restMiddleware? Other files don't call out a function name in module.exports and we aren't calling the function specifically.

lib/web-server.js
```diff
@@ -1,278 +1,271 @@
...
-                server.use(a);  // a = callback
 
-                return server;
-            });
+        /**
+         * Sets up middleware functionality for the web server
+         *
+         * @param {(string|Function)} a
+         * @param {Function} b
+         * @return {this}
```
AbsentSemicolon: Should this be `@return {WebServer} this` since it's returning the class?

lib/web-server.js
```diff
@@ -1,278 +1,271 @@
...
-        return this;
-    }
 
+        /**
+         * This is used for setting up a verb to be called within the application
+         *
+         * @param {string} method
+         * @param {string} route
+         * @param {Function} callback
+         * @return {this}
```
AbsentSemicolon: See above comment.

lib/web-server.js
```diff
@@ -1,278 +1,271 @@
...
-        config = null;  // For safety - don't use this object any longer
-
-        // Make sure our baseUrl does not end with a slash
-        if (this.config.baseUrl.substr(-1) === "/") {
-            this.config.baseUrl = this.config.baseUrl.substr(0, this.config.baseUrl.length - 1);
+        /**
+         * Sets configuration options.
+         *
+         * @param {Object} config
+         * @return this
```
AbsentSemicolon: Should this document `@return {WebServer}`?

AbsentSemicolon:
Looks good to me.

PR #49, opened 2016-04-27
Adding JSON Schema Validator

AbsentSemicolon:


lib/schema.js
```diff
@@ -0,0 +1,85 @@
...
+        return fs.readdirAsync(path).then((files) => {
+            var i, getSchemas;
+
+            getSchemas = [];
+
+            for (i = 0; i < files.length; i += 1) {
+                getSchemas.push(loadSchemaAsync(path + files[i]));
+            }
+
+            return promise.all(getSchemas);
```
fidian: How about using ES6?

```
return fs.readdirAsync(path).then((files) => {
    return promise.all(files.map(loadSchemaAsync))
});
```

spec/helpers/fail-async.helper.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Handles the promise fail.
+ *
+ * @param {string} actual
+ * @param {string} expected
+ * @param {Function} done
+ */
+jasmine.failAsync = function (actual, expected, done) {
+    jasmine.fail(actual, expected);
+    done();
```
fidian: I've rethought this.  Sorry.  This doesn't need to exist.

spec/helpers/test-promise-failure.helper.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {Function} done
+ */
+jasmine.testPromiseFailure = function (promise, errMessage, done) {
+    if (typeof errMessage === "function") {
+        done = errMessage;
+        errMessage = null;
+    }
+
+    return promise.then(function () {
+        jasmine.failAsync("the promise", "rejected", done);
```
fidian: This is the bit I rethought.  Replace this line.

```
done(new Error("The promise should have been rejected"));
```

spec/schema.spec.js
```diff
@@ -0,0 +1,77 @@
...
+        nodeValidator = require("validator");
+        promiseMock = require("./mock/promise-mock");
+        tv4 = require("tv4");
+        schema = require("../lib/schema")(fs, nodeValidator, promiseMock, tv4);
+    });
+    describe(".loadSchemaAsync()", () => {
+        beforeEach(() => {
+            tv4.addSchema = jasmine.createSpy("tv4.addSchema").andCallThrough();
+        });
+        it("loads a schema and validates against it", (done) => {
```
fidian: This should be moved to a separate `describe()`.  I can see us having additional formats validated and the guts of this really is validating against the email format.

spec/schema.spec.js
```diff
@@ -0,0 +1,77 @@
...
+                    ]);
+                }
+
+                return promiseMock.reject("Invalid folder: " + fn.toString());
+            });
+            schema.loadSchemaFolderAsync("/folder/").then(() => {
+                expect(schema.validate("someone@example.net", "email")).toBe(true);
+                expect(schema.validate("someone", "email")).toBe(false);
+                expect(schema.validate(5, "number")).toBe(true);
+                expect(schema.validate(2, "number")).toBe(false);
```
fidian: These validate the validator more than the schema finder.  You don't need negative cases as much here, though what happens when you run `schema.validate({}, "whatever")`?

fidian:
Reviewed.

fidian:
Would you add a `schemas = new Map()` and track what `schema.id`s you have added, then instead of `getSchema` you can use `if (!schemas[schemaId])` and throw based on that.

AbsentSemicolon:
Changed to use a map and using `if (! schemas.has(schema))` to check.

fidian:
Very nice.

PR #50, opened 2016-04-27
Method for encoding data for use in a URL

AbsentSemicolon:
We needed to strip out the +, / and equals base64 can use in it's encoding

### DIFF
--- a/spec/base64.spec.js
+++ b/spec/base64.spec.js
@@ -33,26 +33,26 @@ describe("base64", () => {
             name: "DEAD BEEF"
         }
     ].forEach((scenario) => {
-        it("string encodes " + scenario.name, () => {
+        it("encodes a string to a string: " + scenario.name, () => {
             var result;
 
             result = base64.encode(scenario.decoded);
             expect(result).toEqual(scenario.encoded);
         });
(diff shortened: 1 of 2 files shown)


spec/base64.spec.js
```diff
@@ -60,4 +60,11 @@ describe("base64", () => {
             expect(result.toString("binary")).toEqual(scenario.decoded);
         });
     });
+    it("data encodes for URI", () => {
```
fidian: Grammars it does not.

spec/base64.spec.js
```diff
@@ -60,4 +60,11 @@ describe("base64", () => {
...
+        var result;
+
+        result = base64.encodeForUri(new Buffer("p5>T44d3?12Ui", "binary"));
+        expect(result).toEqual(jasmine.any(Buffer));
+        expect(result.toString()).toBe("cDU-VDQ0ZDM_MTJVaQ");
```
fidian: Always use some encoding when going to string.  In this case, technically it does not matter, though for consistency we should typically use "binary".  This one is not treated as "binary", so use "ascii" for 7-bit characters (which this is).

fidian:
Grammar has been betterfied!  Approvings.

PR #51, opened 2016-04-28
Changing Password to be Secure Hash

AbsentSemicolon:
The password file has absolutely nothing to do with passwords so it makes more sense to change this. Plus this keeps it in line with how we are doing hashes else where in the application.


doc/secure-hash.md
```diff
@@ -0,0 +1,10 @@
+Secure Hash
```
fidian: This file should not exist.

lib/account/account-service.js
```diff
@@ -23,9 +36,10 @@ class AccountService {
      * @param {Object} password
      * @param {Object} Storage
      */
-    constructor(config, password, Storage) {
-        this.password = password;
-        this.storage = Storage.configure(config.storage);
+    constructor(config, secureHash, Storage) {
+        this.config = config;
```
fidian: The directory name should be set to a property here.

```
var accountConfig;

accountConfig = config.account || {};
this.directoryName = accountConfig.accountDir || "account/";
```

lib/account/account-service.js
```diff
@@ -69,13 +83,19 @@ class AccountService {
...
+     * passed in from the manager to direct where to get/put account information.
      *
      * @param {string} accountId
-     * @return {string} the directory
+     * @param {Object} config
+     * @return {Promise.<string>}
      */
-    getDirectory(accountId) {
-        return "account/" + this.password.hashContent(accountId);
+    getDirectoryAsync(accountId, config) {
```
fidian: This should not take a config object here.

lib/secure-hash.js
```diff
@@ -0,0 +1,39 @@
...
+
+        if (! data) {
+            throw new Error("Nothing to hash");
+        }
+
+        if (typeof data === "string") {
+            data = new Buffer(data, "binary");
+        }
+
+        config = config ? config : {};
```
fidian: You did this again.  I would prefer to not use ternary operators.

```
config = config || {};
```

spec/account/account-manager.spec.js
```diff
@@ -82,7 +82,8 @@ describe("AccountManager", () => {
                 previousMfa: "098454",
                 password: "3439gajs933098fj3jfj90aj09fj9390a9023"
             }).then((result) => {
-                expect(result).toEqual({accountId: "aeifFeight3ighrFieigheilw5lfiek",
+                expect(result).toEqual({
+                accountId: "aeifFeight3ighrFieigheilw5lfiek",
```
fidian: Do not pass back the accountId, currentMfa, previousMfa, password.  In fact, don't return anything besides the accountIdHash that they will use for future API calls.  I understand that this is just returning what the mock returns, but the manager shouldn't need to expose this stuff.

spec/secure-hash.spec.js
```diff
@@ -0,0 +1,44 @@
...
+            hashLength: 24,
+            iterations: 10000,
+            salt: "pinkFluffyUnicornsDancingOnRainbows"
+        }).then((result) => {
+            expect(result.toString("binary")).toBe("9GnOLZ_xAlfMA4C6DHsjNJJpsShI_TgR");
+        }).then(done, done);
+    });
+    it("hashes without a config being passed in", (done) => {
+        secureHash.hashAsync("rRTcBER_EiFUsRa34Hj5Zpok").then((result) => {
+            expect(result.toString("binary").length).toBe(683);
```
fidian: What should this result be?  We're not testing the hashed value here, though the defaults should be known and never change.

lib/secure-hash.js
```diff
@@ -0,0 +1,39 @@
...
+module.exports = function (base64, crypto, promise) {
+    crypto.pbkdf2Async = promise.promisify(crypto.pbkdf2);
+
+    /**
+     * Hashes the content passed using defaults or
+     * configuration options if passed in.
+     *
+     * @param {(Buffer|string)} data
+     * @param {Object} config
+     * @return {Promise.<Buffer>}
```
fidian: This is not right.  It returns a URI-safe string, not a buffer.

spec/secure-hash.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        secureHash = require("../lib/secure-hash")(base64, crypto, promiseMock);
+    });
+    it("hashes a passed in string", (done) => {
+        secureHash.hashAsync("rRTcBER_EiFUsRa34Hj5Zpok", {
+            algorithm: "sha256",
+            hashLength: 24,
+            iterations: 10000,
+            salt: "pinkFluffyUnicornsDancingOnRainbows"
+        }).then((result) => {
+            expect(result.toString("binary")).toBe("9GnOLZ_xAlfMA4C6DHsjNJJpsShI_TgR");
```
fidian: Here, and elsewhere in the file, you convert a string to a string.  That is unnecessary.

lib/secure-hash.js
AbsentSemicolon: Ah yes, I meant to change this to the way above.  Missed that before I sent the pull request.

doc/account.md
```diff
@@ -8,10 +8,23 @@ The `account` options in `config.json` allow you to set how long an account is v
...
     "account": {
+        "accountDir": "account/",
         "completeLifetime": {
             "months": 6
         },
+        "idHash": {
+            "algo": "sha256",
+            "hashLength": 24,
+            "iterations": 10000,
+            "salt": "pinkFluffyUnicornsDancingOnRainbows"
```
fidian: Would you change this to something that's not as embarrassing if I show my Uncle Mark (a PCI expert)?

lib/secure-hash.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = function (base64, crypto, promise) {
+    crypto.pbkdf2Async = promise.promisify(crypto.pbkdf2);
+
+    /**
+     * Hashes the content passed using defaults or
+     * configuration options if passed in.
+     *
+     * @param {(Buffer|string)} data
+     * @param {Object} config
```
fidian: Should document this in a typedef.

fidian: It also looks optional, thus the documentation should be something like

```
@param {secureHash~config} [config]
```

spec/account/account-service.spec.js
```diff
@@ -25,31 +25,40 @@ describe("AccountService", () => {
...
             });
             accountService.completeAsync("directory", {
                 password: "somereallylonghashedpassword"
             }, {
                 expires: new Date()
             }).then((result) => {
                 expect(result).toEqual({
-                    email: "some.one@example.net",
-                    password: "somereallylonghashedpassword"
+                    accountId: "unhashedAccountId"
```
fidian: This should be returning the hashed account ID, not the unhashed.  We supplied the unhashed one, yes?  Or pulled it from the storage mechanism?

lib/secure-hash.js
```diff
@@ -0,0 +1,39 @@
...
+            throw new Error("Nothing to hash");
+        }
+
+        if (typeof data === "string") {
+            data = new Buffer(data, "binary");
+        }
+
+        config = config || {};
+        algorithm = config.algo || "sha512";
+        hashLength = config.hashLength || 512;
```
fidian: Lower this considerably.  How about 48 for a total of 64 characters once encoded?

4 * bytesRaw / 3 = bytesEncoded.  33.33333333333333333% (roughly) increase in size.

fidian:
Approved!

PR #52, opened 2016-04-28
Add CORS middleware

fidian:
This adds `restify.CORS()` to the list of middleware for standard handling of the options calls that HTTP clients will make during preflight requests.

The tests are also improved so they not only ensure that the middleware was called but also that it was added to the server via `server.use()`.

Finally, there is a change that updates the team name to try to prevent confusion with another GitHub organization.

### DIFF
--- a/spec/rest-middleware.spec.js
+++ b/spec/rest-middleware.spec.js
@@ -16,8 +16,8 @@ describe("restMiddleware", () => {
             helmetMock.noCache,
             helmetMock.noSniff,
             helmetMock.xssFilter,
-            restifyMock.CORS,
-            restifyLinks
+            restifyLinks,
+            restifyMock.CORS
         ].forEach((spy) => {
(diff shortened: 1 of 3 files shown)


spec/rest-middleware.spec.js
```diff
@@ -1,30 +1,46 @@
...
-        expect(restifyLinks).toHaveBeenCalled();
+        [
+            helmetMock.frameguard,
+            helmetMock.ieNoOpen,
+            helmetMock.hidePoweredBy,
+            helmetMock.noCache,
+            helmetMock.noSniff,
+            helmetMock.xssFilter,
+            restifyMock.CORS,
+            restifyLinks
```
AbsentSemicolon: Doesn't L come before M?

AbsentSemicolon:
Reviewed and looks good.

PR #55, opened 2016-04-28
Adding restify-router-magic for automatic route generation

fidian:
**NOTE** *Do not approve until after reviewing [restify-router-magic](https://github.com/tests-always-included/restify-router-magic)!*

Added the third party library.

Changed `WebServer.addRoute()` to `WebServer.addRoutes(folder)` and
that hooks into restify-router-magic.

Changed ApiServer from a class into an exported function.  Calling the
function starts the server.

Added a shebang and changed the permissions of `bin/server.js` so it can
start up from the command line.

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -1,6 +1,8 @@
+#!/usr/bin/env node
+
 "use strict";
 
-var ApiServer, config, container, logger;
+var config, container, logger;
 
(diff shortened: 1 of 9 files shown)


AbsentSemicolon:
Reviewed and looks good.

PR #56, opened 2016-04-29
Secure hash comparison

AbsentSemicolon:
This checks hashes are the same. Tests take some time to run. In c9.io it took 8 minutes to complete a suite of tests, but could be faster as these machines don't seem to be as fast as they could be.


debug/test-hash-compare.js
```diff
@@ -0,0 +1,63 @@
...
+
+        // run bad
+        start = process.hrtime();
+        while((process.hrtime()[0] - start[0]) < timeToRunFor) {
+            secureHash.compare(against, bad);
+            badTimes += 1;
+        }
+
+        console.log("   Times Bad Hashes Ran: ", badTimes);
+
```
fidian: Did not add a test for the off-by-one buffer length.  We want to make sure that we don't leak timing information when buffer lengths do not match.

lib/secure-hash.js
```diff
@@ -19,6 +19,33 @@ module.exports = function (base64, crypto, promise) {
     crypto.pbkdf2Async = promise.promisify(crypto.pbkdf2);
 
     /**
+     * Compares a hashed string to another hashed string
+     * to see if they are the same. We loop through the
+     * shortest one to see what compares to the longer one.
+     *
+     * We want to always loop through the data ever if we know
```
fidian: s/ever/even/

lib/secure-hash.js
```diff
@@ -19,6 +19,33 @@ module.exports = function (base64, crypto, promise) {
...
+     * the lengths are different. This helps to deter against
+     * time attacks.
```
fidian: picosecond timing attacks

fidian:
Extremely minor notes on the code review.

fidian:
Looks great!

PR #58, opened 2016-05-02
Account creation using new "registration" flow

AbsentSemicolon:
Accounts will now create a registration file with the information needed to create an account which will be used until the account holder/client/user goes through the complete process to create their account.

After the account is made the registration file will be deleted. This helps to keep things cleaner in the account directory as we won't have half finished files sitting around.


config.json
```diff
@@ -12,7 +12,8 @@
         },
         "initiateLifetime": {
             "hours": 1
-        }
+        },
+        "registrationDir": "registration/"
```
fidian: Needs documentation.

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
 "use strict";
-/**
- * Manages the business logic of account controls.
- */
 
 /**
- * The object sent in which has information needed to validate the user
+ * The object sent container information needed to validate the user
```
fidian: I don't get this.  The object sent container information?  What is container information?

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
...
  * has properly set up their two factor authentication and has generated
- * a password for use to save.
+ * a password for use.
  *
  * @typedef {Account~accountinfo}
- * @param {string} accountId this is the unhashed account id sent to client
- * @param {string} currentMfa
- * @param {string} previousMfa
- * @param {string} password
+ * @property {string} accountId this is the unhashed account id sent to client
```
fidian: You don't ever need to say "this is" for the descriptions.  You really don't need "the" either.

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
...
+ * @property {string} password
  */
-class AccountManager {
+
+/**
+ * This is what is sent back to account holder to confirm their
+ * sign up by showing the password salt and mfa key we've
+ * generated for them.
+ *
+ * @typedef {Account~confirm}
```
fidian: The typedef should be `accountManager~confirm` instead of `Account~confirm`.

* accountManager is not a class.
* There is no Account class.

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
...
- */
 
 /**
- * The object sent in which has information needed to validate the user
+ * The object sent container information needed to validate the user
  * has properly set up their two factor authentication and has generated
- * a password for use to save.
+ * a password for use.
  *
  * @typedef {Account~accountinfo}
```
fidian: Change this to `accountManager~accountInfo`

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
...
+
+
+    /**
+     * Finishes setting up the account and calls the Account Service
+     * to save an account file with the requested information.
+     *
+     *
+     * @param {string} accountId
+     * @param {Object} accountInfo
+     * @return {Promise.<Object>}
```
fidian: What type of Object is provided by the promise?

lib/account/account-manager.js
```diff
@@ -1,36 +1,80 @@
...
+    }
+
+
+    /**
+     * Finishes setting up the account and calls the Account Service
+     * to save an account file with the requested information.
+     *
+     *
+     * @param {string} accountId
+     * @param {Object} accountInfo
```
fidian: What type of object is the accountInfo parameter?

lib/account/account-manager.js
```diff
@@ -43,60 +87,67 @@ class AccountManager {
      * hash(hash(password) + salt)
      *
      * @param {AccountManager~accountInfo} accountInfo
-     * @return {Promise.<AccountService~accountFile>}
+     * @return {Promise.<Object>} the accountId is sent back to the user when account file is made
```
fidian: What type of object?  This should match `setupAccountAsync`.

lib/account/account-manager.js
```diff
@@ -43,60 +87,67 @@ class AccountManager {
...
+    }
 
-                options = {
-                    expires: this.otDate.now().plus(this.config.account.completeLifetime)
-                };
 
-                return this.accountService.completeAsync(directory, accountInfo, options);
-            });
+    /**
+     * Confirms the account holder wants to confirm the account and get the information
```
fidian: Confirms the confirmation?

lib/account/account-manager.js
```diff
@@ -43,60 +87,67 @@ class AccountManager {
...
-    initiateAsync(accountInfo) {
-        return this.promise.props({
-            accountId: this.random.password(24),
-            mfaKey: this.hotp.generateSecretAsync(),
-            salt: this.random.password(128)
+    function signupInitiationAsync(accountInfo) {
+        return promise.props({
+            mfaKey: hotp.generateSecretAsync(),
+            regId: random.passwordAsync(24),
+            passwordSalt: random.passwordAsync(128)
```
fidian: Rename `random.passwordAsync` since you are not generating passwords.  You're generating identifiers or random encoded strings.

lib/account/account-service.js
```diff
@@ -1,99 +1,140 @@
 "use strict";
 
 /**
- * What we initially send back to the client so they can start to initiate
- * the complete procedure and save information.
+ * What we initially generate in the registration part of storage
```
fidian: Please reword.  Remove "what we initially" stuff.  Focus more on describing the idea of the object instead of how we use it.

lib/account/account-service.js
```diff
@@ -1,99 +1,140 @@
...
  * @property {string} mfaKey
- * @property {string} salt
+ * @property {string} passwordSalt
  */
 
 /**
- * Stored information for the account.  This is changed from the initiated
+ * Stored information for the account. This is changed from the initiated
  * file as we add password and there could be other items added after
  * we initiate the account.
```
fidian: Same comment here.  This is a finalized account record.  Try to not focus on what happens to it.

lib/account/account-service.js
```diff
@@ -1,99 +1,140 @@
...
+                        };
+                    });
+                });
             });
         });
     }
 
 
     /**
+     * Creats the account file in the account directory.
```
fidian: Missing an "e" in Creates.

lib/account/account-service.js
```diff
@@ -1,99 +1,140 @@
...
      *
      * TODO: Decrypt data
      *
-     * @param {string} accountId
+     * @param {string} directory
      * @return {Promise.<AccountService~accountFile>}
      */
-    getAsync(directory) {
-        return this.storage.getAsync(directory).then((accountFile) => {
+    function getAsync (directory) {
```
fidian: No space before the open parenthesis.

lib/account/account-service.js
```diff
@@ -1,99 +1,140 @@
...
      *
-     * @param {string} accountId
+     * @param {string} prefix
+     * @param {string} idToHash
      * @return {Promise.<string>}
      */
-    getDirectoryAsync(accountId) {
-        return this.secureHash.hashAsync(accountId, this.accountConfig.idHash || {}).then((hashedAccountId) => {
-            return this.directoryName + hashedAccountId;
+    function getDirectoryAsync (prefix, idToHash) {
```
fidian: No space before the open parenthesis.  (not mentioning it again, though it is in other spots)

config.json
AbsentSemicolon: Documented this and updated with the new way we create accounts.

lib/account/account-manager.js
AbsentSemicolon: Should have been contains...

lib/account/account-service.js
AbsentSemicolon: I'll have to look into getting rule in eslint sometime to make sure I don't keep doing that.

lib/account/account-service.js
```diff
@@ -1,99 +1,128 @@
...
-                return {
-                    accountId: accountFile.accountId
-                };
+                return makeAccountFileAsync(accountFile, options).then(() => {
+                    return storage.delAsync(regDirectory).then(() => {
+                        return {
+                            accountId: accountInfo.accountId
+                        };
+                    });
+                });
```
fidian: I would rather see this inside the function:

```
return getDirectoryAsync(registrationDirName, regId).then(getAsync).then((accountFile) => {
```

lib/account/account-service.js
```diff
@@ -1,99 +1,128 @@
...
+     * Get the registration file when it's still in the registration directory.
+     * This is to eliminate one steps for the manager getting the directory
+     * and then requesting the file.
+     *
+     * @param {string} regId
+     * @return {Promise.<accoutService~accountFile>}
+     */
+    function getRegistrationFileAsync(regId) {
+        return getDirectoryAsync(registrationDirName, regId).then((directory) => {
+            return getAsync(directory).then((accountFile) => {
```
fidian: Same here as above.

```
return getDirectoryAsync(registrationDirName, regId).then(getAsync);
```

Also the one callback that just returns the value that was passed in is unneeded.

lib/account/account-service.js
```diff
@@ -1,99 +1,128 @@
...
-     * @param {string} accountId
+     * @param {string} prefix
+     * @param {string} idToHash
      * @return {Promise.<string>}
      */
-    getDirectoryAsync(accountId) {
-        return this.secureHash.hashAsync(accountId, this.accountConfig.idHash || {}).then((hashedAccountId) => {
-            return this.directoryName + hashedAccountId;
+    function getDirectoryAsync(prefix, idToHash) {
+        return secureHash.hashAsync(idToHash, accountConfig.idHash || {}).then((hashedId) => {
```
fidian: The `|| {}` is unnecessary.

lib/account/account-service.js
```diff
@@ -104,16 +133,23 @@ class AccountService {
...
      * @param {Object} options
-     * @return {Promise.<AccountService~initiateResult>}
+     * @return {Promise.<accountManager~initiationInfo>}
      */
-    initiateAsync(directory, accountInfo, options) {
-        return this.storage.putAsync(directory, JSON.stringify(accountInfo), options).then(() => {
-            return accountInfo;
+    function signupInitiateAsync(accountInfo, options) {
+        return getDirectoryAsync(registrationDirName, accountInfo.regId).then((directory) => {
+            return storage.putAsync(directory, JSON.stringify(accountInfo), options).then(() => {
```
fidian: I would prefer you put the next `.then()` outside.

```
return getDirectoryAsync(registrationDirName, accountInfo.regId).then((directory) => {
    return storage.putAsync(directory, JSON.stringify(accountInfo), options);
}).then(() => {
    return {
        regId.accountInfo.regId;
    };
});
```

This will help prevent our code from becoming terribly indented.

lib/random.js
```diff
@@ -24,14 +24,14 @@ module.exports = function (base64, crypto, promise) {
...
-         * Generates a random password fulfilling the size requirements.
+         * Generates a random string fulfilling the size requirements.
          * Does this by generating sufficient random data, converting it
          * to Base64, then doing some string manipulation.
          *
          * @param {number} size
          * @return {Promise.<string>}
          */
-        passwordAsync: function (size) {
+        randomizedAsync: function (size) {
```
fidian: Rename to something that is more aptly describing what this does or what it makes.

```
randomIdAsync: function (size) {
```

doc/account.md
```diff
@@ -44,16 +44,16 @@ Initiating an Account
 
 The first action for an account is `initiate`. This sets up a few pieces of data needed to properly generate and track an account. We generate a `regId`, `mfaKey`, and `passwordSalt`. An email address must be passed in. These are then passed back to the user to store on their own and will be used to generate their password.
 
-This will save the information to a file named after the hash of the `regId` in the `registrationDir` location. This registration file will be deleted upons successfully creating and account.
+This will save the information to a file named after the hash of the `regId` in the `registrationDir` location. This registration file will be deleted upons successfully creating an account.
```
fidian: No mention of storage.  Storage should get documented as well, plus additional links scattered in to make sure people know we are not talking about a local disk.

Imagine someone only reading this paragraph instead of the whole doc.  Everyone skims, so try to make each sentence unambiguous and clear.

doc/account.md
```diff
@@ -44,16 +44,16 @@ Initiating an Account
...
+The next action is to confirm an account. This requires the user to access the URL using the `regId` we gave them. We will then compare it's valid by comparing the hashes then return the `mfaKey` and `passwordSalt` so they can proceed to the next step.
 
 Completing an Account
 ---------------------
 
-To complte an account, the client will need to pass in a hashed `password`, the `regId` which we gave them previously as well a `currentMfa` code and `previousMfa` code. This will allow us to find the registration file using the `regId` hashing it and looking it up in storage. Getting the `mfaKey` off the account we are able to then validate the user using the MFA codes they passed in.
+To complete an account, the client will need to pass in a hashed `password`, the `regId` which we gave them previously as well a `currentMfa` code and `previousMfa` code. This will allow us to find the registration file using the hashed `regId` and looking it up in storage. Getting the `mfaKey` off the account we are able to then validate the user using the MFA codes they passed in.
```
fidian: "as well a" should be "as well as a"

doc/account.md
```diff
@@ -44,16 +44,16 @@ Initiating an Account
...
 
-If the account validates, we create the account file from the registration file, remove the `regId` as it's no longer needed, added the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
\ No newline at end of file
+If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
```
fidian: If I only read this line, it looks like it is saved to disk, not to s3.

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
 
 /**
- * The object sent in which has information needed to validate the user
- * has properly set up their two factor authentication and has generated
- * a password for use to save.
+ * Contains information needed to validate the user has
+ * properly set up their two factor authentication and
+ * has generated a password.
+ *
+ * @typedef {Object} accountManager~accountinfo
```
fidian: Typedef seems wrong to me.  Just the "accountinfo" bit (assuming you meant "accountInfo").  This is not an account, nor is it information about an account.  It is perhaps `accountManager~registrationInfo`.

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
+ * @property {string} regId Unhashed registration id sent to client
+ * @property {string} currentMfa
+ * @property {string} previousMfa
+ * @property {string} password
```
fidian: Shouldn't this also contain email?  We're not sending an email, but we should still accept one at this stage.

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
+ * @property {string} mfaKey
  */
-class AccountManager {
+
+/**
+ * Contains the registration id needed for a user to
+ * continue on with the sign up process.
+ *
+ * @typeDef {Object} accountManager~completionInfo
+ * @property {string} accountId Unhased account id generated
```
fidian: "unhased" -> "unhashed" (more than one appearance in these files - search across all files)

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
+ */
+
+/**
+ * Contains the accountId an account holder needs to
+ * use the service.
+ *
+ * @typeDef {Object} accountManager~initiationInfo
```
fidian: How does this differ from `accountManager~initiationInfo`

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
+
+/**
+ * @param {Object} accoutService
+ * @param {Object} config
+ * @param {Object} hotp
+ * @param {Object} otDate
+ * @param {Object} random
+ * @param {Object} promise
+ * @return {Object} account manager object
+ */
```
fidian: Remove this doc block per our earlier discussion.

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
+        var options, regId;
+
+        accountInfo.accountId = accountId;
+        accountInfo.password = accountInfo.password;
+        regId = accountInfo.regId;
+        options = {
+            expires: otDate.now().plus(config.account.completeLifetime)
+        };
+
+        return accountService.completeAsync(accountInfo, options, regId);
```
fidian: I would feel better if we didn't simply return everything the accountService returns and instead map desired values to the output, if we want to preserve anything at all.  That would make exposure of information a deliberate act instead of something that could be accidentally done.

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
 
 
     /**
      * Validates the previous and current MFA values are valid and sends
-     * data off to Account service class for updating/putting data into
-     * the Account record.
+     * data off to the account service for updating/putting data into
+     * the account record.
      *
      * Password generation should follow this format:
```
fidian: This bit of very useful information advice should be in the documentation (doc/) instead of buried in code.

What hashing mechanism is used?  If it is configurable, that should also be in the documentation.  As a hint, we may wish to supply the hashing method in the response for the user.

Consider:  Should we replace this hashing with pbkdf2 so it is more secure?

lib/account/account-manager.js
```diff
@@ -1,102 +1,171 @@
...
 
-                options = {
-                    expires: this.otDate.now().plus(this.config.account.completeLifetime)
-                };
 
-                return this.accountService.completeAsync(directory, accountInfo, options);
-            });
+    /**
+     * Gets the account file from the registration id passed in
+     * so the account holder can finished signing up.
```
fidian: "can finished" -> "can finish"

lib/account/account-service.js
```diff
@@ -1,70 +1,74 @@
...
+ * Handles the creating, deletion, and updating of account and
+ * registration files.
  */
 
 /**
- * Stored information for the account.  This is changed from the initiated
- * file as we add password and there could be other items added after
- * we initiate the account.
+ * Information about an account holder. Used to generate/validate
+ * a password and two factor authentication as well as Ids and
```
fidian: Ids is not a proper noun and thus should be lowercase or spelled out as "identifiers"

lib/account/account-service.js
```diff
@@ -1,70 +1,74 @@
...
+ * contact information.
+ *
+ * Once the account is complete, regId is removed.
```
fidian: regId is not in this object, so this comment should be removed.

lib/account/account-service.js
```diff
@@ -1,70 +1,74 @@
...
+ * @property {string} passwordSalt
  */
 
 /**
- * Handles account creation from initiating to confirmation.
+ * @param {Object} config
+ * @param {Object} secureHash
+ * @param {Object} Storage
+ * @return {Object} account service object
  */
```
fidian: Drop this doc block per our previous comment.

lib/account/account-service.js
```diff
@@ -1,70 +1,74 @@
...
-        this.secureHash = secureHash;
-        this.storage = Storage.configure(config.storage);
-    }
+module.exports = function (config, secureHash, Storage) {
+    var accountConfig, accountDirName, registrationDirName, storage;
 
+    accountConfig = config.account || {};
+    accountDirName = accountConfig.accountDir || "account/";
+    registrationDirName = accountConfig.registrationDir || "registration/";
+    storage = new Storage().configure(config.storage);
```
fidian: Shouldn't we have "storage" injected instead of "Storage"?

Please consider this and maybe we can come up with something.

lib/account/account-service.js
```diff
@@ -75,45 +79,69 @@ class AccountService {
...
+    function getAsync(directory) {
+        return storage.getAsync(directory).then((accountFile) => {
             return JSON.parse(accountFile.toString("binary"));
         });
     }
 
 
     /**
+     * Get the registration file when it's still in the registration directory.
+     * This is to eliminate one steps for the manager getting the directory
```
fidian: Eliminate one steps -> eliminate one step

lib/mfa/hotp.js
```diff
@@ -4,36 +4,34 @@
...
  * application we want to verify the client has access to.
  *
+ * We want to promisfy all methods we can because some of them
+ * take some time to generate and we don't want code going on
+ * till we have the data we need back.
+ *
+ * @param {Object} config
+ * @param {Object} twofa
+ * @param {Object} promise
+ * @return {Object} hotp object
```
fidian: The words on top are great.  The explanation of the paramters are pretty much not great.  Please remove per our earlier discussion.

spec/account/account-manager.spec.js
```diff
@@ -57,69 +38,114 @@ describe("AccountManager", () => {
...
                 });
-                expect(result.accountId.length).toBe(24);
-                expect(result.salt.length).toBe(128);
+                expect(result.regId.length).toBe(24);
+            }).then(done, done);
+        });
+        it("gets the registration id back using config options", (done) => {
+            var accountManagerLocal;
 
+            accountManagerLocal = require("../../lib/account/account-manager")(accountServiceFake, {
```
fidian: You are changing config options and making a lot of lines duplicated in your test.  Instead of exposing a bunch of mocks, why not expose a function to make the manager?  You can put this in your beforeEach().

```
create = (config) => {
    return require("../../lib/account/account-manager")(accountServiceFake, config || defaultConfig, hotpFake, otDateMock, promiseMock, randomMock);
}
```

Remove the creation of accountManager in the beforeEach().  Then just call `accountManager = create({...})` in your tests.

spec/account/account-service.spec.js
```diff
@@ -1,32 +1,44 @@
...
-        });
-        storageFake.getAsync.andCallFake(() => {
-            return promiseMock.resolve(
-                new Buffer('{"data": "thing"}', "binary")
-            );
-        });
-        storageFake.putAsync.andCallFake(() => {
-            return promiseMock.resolve(true);
-        });
+        StorageFake = class StorageFake {
```
fidian: Make this a mock that other things can load.

fidian: Remember to consider what we should do to make this an instance instead of a class.

fidian:
Reviewed.

lib/account/account-manager.js
AbsentSemicolon: We have the email already in the account file. We don't need one here, this is to give the password and complete the account creation.

doc/account.md
```diff
@@ -12,7 +12,7 @@ The `idHash` property of `account` covers how the ids will be hashed when stored
 
 You can find a list of available hashing methods in `lib/ciphers-and-hashes.js`.
 
-Where the account information is stored can be change by updating `accountDir`. The registration files can be updated by changing `registrationDir`. Ideally these would be different as we store temporary registration files during the sign up process in the `registrationDir`, and permanent files for the account in `accountDir` once sign up is complete. These cannot be local directories as they will be passed to the storage mechanism to handle.
+Where the account information is stored can be change by updating `accountDir`. The registration files can be updated by changing `registrationDir`. Ideally these would be different as we store temporary registration files during the sign up process in the `registrationDir`, and permanent files for the account in `accountDir` once sign up is complete. These cannot be local directories as they will be passed to the storage engine to handle.
```
fidian: link `[storage engine](storage-engine.md)`

fidian: Link in several places.  Anywhere storage is mentioned.

doc/account.md
```diff
@@ -56,4 +56,23 @@ Completing an Account
...
 
-If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
\ No newline at end of file
+If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
+
+Passwords
+---------
+
+Passwords are handled on the clients side. We should never see an unhashed password, unless of course the client sends one in to use, but we won't store that anywhere except their encrpyted account file.
```
fidian: s/encrpyted/encrypted/

Don't mention unhashed.  Instead we should focus on what we do support - we only deal with hashed passwords.

doc/account.md
```diff
@@ -56,4 +56,23 @@ Completing an Account
...
+    hash(password + passwordSalt)
+```
+
+When sending the password in for authentication, the password should be hashed again and then hashed with the challenge the applictation sent them.
+
+```
+    hash(hash(hashedPassword) + challenge)
+```
+
+The hashing of the password is primarily up to the client, as we do nothing with the password except save it in their account file. But the hashing of the challenge and then subsequent hashing after that will be set in `config.json` and will utilize `pbkdf2` on both sides to maintain a consistent login paradigm.
```
fidian: This is incorrect.

We dictate exactly how the hashing should happen, though it should be in a configuration file.

doc/account.md
```diff
@@ -56,4 +56,23 @@ Completing an Account
...
+Passwords should be hashed along with the `passwordSalt` we provided them during account signup.
+
+```
+    hash(password + passwordSalt)
+```
+
+When sending the password in for authentication, the password should be hashed again and then hashed with the challenge the applictation sent them.
+
+```
+    hash(hash(hashedPassword) + challenge)
```
fidian: Either indent or use code fences.  I prefer no code fences and using just indentation.

Here and above in the file.

doc/storage.md
```diff
@@ -0,0 +1,17 @@
...
+
+To configure the engine, `bucket` will set where the engine will be getting, putting, listing, and deleting the data from, while `region` lets you decide the region the data should be stored in. *These are if you are using Amazon's S3, but the these keys can change if need be.*
+
+```
+    "storage": {
+        "bucket": "opentoken-io-storage",
+        "engine": "s3",
+        "region": "us-east-1"
+    }
+```
```
fidian: Remove code fences.

doc/storage.md
```diff
@@ -0,0 +1,17 @@
+Storage
+=======
+
+Configuration
+-------------
+
+The `storage` options in `config.json` help to handle many facets of how storage is accomplished in OpenToken.io. Setting `engine` to your desired code in the storage directory will let set up the application to be able to inject your desired engine into code where it will be needed.
+
+To configure the engine, `bucket` will set where the engine will be getting, putting, listing, and deleting the data from, while `region` lets you decide the region the data should be stored in. *These are if you are using Amazon's S3, but the these keys can change if need be.*
```
fidian: Split the "s3" bit to a new heading.

```
### S3
```

Also, change "but the these"

lib/account/account-manager.js
```diff
@@ -72,6 +63,36 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
 
 
     /**
+     * Maps the information we specifially want from the
+     * service to what we want to return to the account holder
+     * for continued use of the system.
+     *
+     * @param {Object} details
```
fidian: What type of object is passed in?

lib/account/account-manager.js
```diff
@@ -72,6 +63,36 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+        };
+    }
+
+
+    /**
+     * Maps the information we specifially want from the
+     * service to what we want to return to the account holder
+     * for registration.
+     *
+     * @param {Object} details
```
fidian: What type of details are passed in?

lib/account/account-manager.js
```diff
@@ -160,7 +178,7 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
             };
 
             return accountService.signupInitiateAsync(accountFile, options, bits.regId);
-        });
+        }).then(registrationDetails);
```
fidian: You should update the `@return` in the doc block to be accurate.

lib/account/account-service.js
```diff
@@ -7,31 +7,22 @@
...
  */
 
-/**
- * @param {Object} config
- * @param {Object} secureHash
- * @param {Object} Storage
- * @return {Object} account service object
- */
-module.exports = function (config, secureHash, Storage) {
+module.exports = function (config, secureHash, storage) {  // I would like to use Storage, because that's what we have in the dependencies and it would keep that link in mind.
```
fidian: Um, I think that comment is misplaced in this file.

lib/account/account-service.js
```diff
@@ -49,7 +40,9 @@ module.exports = function (config, secureHash, Storage) {
         return getDirectoryAsync(registrationDirName, regId).then(getAsync).then((accountFile) => {
             accountFile.password = accountInfo.password;
 
-            return makeAccountFileAsync(accountFile, options).then(storage.delAsync);
+            return makeAccountFileAsync(accountInfo.accountId, accountFile, options).then(() => {
+                storage.delAsync;
```
fidian: This is not calling `delAsync()` the function.  It is returning the function.

fidian: You said the tests pass.  This means you should first fix the tests to detect this error and then fix the code.

lib/account/account-service.js
```diff
@@ -65,9 +58,9 @@ module.exports = function (config, secureHash, Storage) {
      * @param {Object} options
      * @return {Promise.<string>} the directory used to put the file
      */
-    function makeAccountFileAsync(accountInfo, options) {
-        return getDirectoryAsync(accountDirName, accountInfo.accountId).then((accountDir) => {
-            return storage.putAsync(accountDir, JSON.stringify(accountInfo), options).then(() => {
+    function makeAccountFileAsync(accountId, accountFile, options) {
+        return getDirectoryAsync(accountDirName, accountId).then((accountDir) => {
+            return storage.putAsync(accountDir, JSON.stringify(accountFile), options).then(() => {
                 return accountDir;
```
fidian: Consider - should this return the account file location entirely, not just the directory?

It looks as though `accountDir` is just a bad name for a variable.  If that's the case, rename the variable and update the doc block.

lib/random.js
```diff
@@ -19,7 +19,7 @@ module.exports = function (base64, crypto, promise) {
          * @param {number} size
          * @return {Promise.<Buffer>}
          */
-        bufferAsync: function (size) {
+        bufferAsync: function(size) {
```
fidian:
```
function () {}
function name() {}
```

With name, no space before `(`.  Without a name, include the space.

lib/storage/s3.js
```diff
@@ -1,142 +1,144 @@
...
+    var awsS3, s3Bucket;
+
+    awsS3 = null;
+    s3Bucket = null;
+    configure(config.storage);
+
+    /**
+     * Configures options for AWS.
+     *
+     * @param {Object} config
```
fidian: Should typedef this so we can document the properties that are used/required.

fidian: Would *greatly* prefer that the config passed in uses a namespaced config.  It makes it much easier to switch between multiple configs.  Here's a sample:

```
config = {
    storage: {
        engine: "s3",
        s3: {
            region: "us-east-1"
        }.
        disk: {
            base: "/var/lib/opentoken/"
        }
    }
}
```

lib/random.js
fidian: You should use ES6 functions here too.

```
bufferAsync: () => {
```

spec/account/account-manager.spec.js
```diff
@@ -51,32 +41,24 @@ describe("AccountManager", () => {
         otDateMock = require("../mock/ot-date-mock");
         promiseMock = require("../mock/promise-mock");
         randomMock = require("../mock/random-mock");
-        accountManager = require("../../lib/account/account-manager")(accountServiceFake, config, hotpFake, otDateMock, promiseMock, randomMock);
+        create = (config) => {
+            return require("../../lib/account/account-manager")(accountServiceFake, config, hotpFake, otDateMock, promiseMock, randomMock);
```
fidian: Consider abbreviating the code by exposing a default configuration that `create` uses.

```
defaultConfig = { ... };
create = () => {
    return require("...")(accountServiceFake, defaultConfig, ...)
};
```

spec/mock/random-mock.js
```diff
@@ -19,6 +19,12 @@ random.bufferAsync.andCallFake((size) => {
 });
 
 random.randomIdAsync.andCallFake((size) => {
+    if (! size) {
```
fidian: Add a comment that the real code doesn't do this.  We're adding this check so our tests can help catch scenarios where size can be zero.

... or make it act like the real class.

Throwing is just fine here too.

spec/storage.spec.js
```diff
@@ -0,0 +1,45 @@
...
+        expect(() => {
+            create({
+                storage: {
+                    engine: "s3"
+                }
+            });
+        }).not.toThrow();
+        expect(containerMock.call).toHaveBeenCalled();
+    });
+    it("has no config options", () => {
```
fidian: Try to word this sentence to describe the behavior you expect.

The storage .....

Your test says this:  The storage has no config options.  That's not behavior.  Try something like "will error without a proper configuration".

This comment applies to the other tests.

spec/storage/s3.spec.js
```diff
@@ -30,63 +33,36 @@ describe("storage/s3", () => {
...
+        }, promiseMock);
     });
-    describe("del()", (done) => {
+    describe(".delAsync()", (done) => {
         it("deletes a file", () => {
             s3.delAsync("afile").then((val) => {
                 expect(val).toEqual({
                     Key: "afile"
                 });
+                expect(awsSdkMock.config.region).toBe("us-east-1");
```
fidian: This assertion should be done in a separate test.  It's not exclusive to `delAsync()`.

doc/account.md
AbsentSemicolon: Don't know why I did it that way since I did it without the fences before.

doc/storage.md
AbsentSemicolon: Again, don't know why I did it that way since I did it without the fences before.

doc/storage.md
AbsentSemicolon: Split this off to explain better about the configuration options for S3.

lib/random.js
AbsentSemicolon: Using ES6 closures now. Makes more sense.

lib/storage/s3.js
AbsentSemicolon: Split up the options to a separate area of storage configurations.

spec/storage.spec.js
AbsentSemicolon: Updated these to make more sense hopefully.

spec/storage/s3.spec.js
AbsentSemicolon: Exposed transit so we can call it in the tests, but transit isn't used outside of the module.

doc/account.md
```diff
@@ -61,18 +61,14 @@ If the account validates, we create the account file from the registration file,
 Passwords
```
fidian: Link up the "storage" on line 57.

doc/account.md
```diff
@@ -61,18 +61,14 @@ If the account validates, we create the account file from the registration file,
...
 ---------
 
-Passwords are handled on the clients side. We should never see an unhashed password, unless of course the client sends one in to use, but we won't store that anywhere except their encrpyted account file.
+Passwords are handled on the clients side and then stored their encrypted account file upon completing the sign up procss.
```
fidian: s/clients/client's/

doc/account.md
```diff
@@ -61,18 +61,14 @@ If the account validates, we create the account file from the registration file,
...
 
 When sending the password in for authentication, the password should be hashed again and then hashed with the challenge the applictation sent them.
 
-```
     hash(hash(hashedPassword) + challenge)
-```
 
-The hashing of the password is primarily up to the client, as we do nothing with the password except save it in their account file. But the hashing of the challenge and then subsequent hashing after that will be set in `config.json` and will utilize `pbkdf2` on both sides to maintain a consistent login paradigm.
\ No newline at end of file
+The hashing of the password is primarily up to the client, as we only store the password. The hashing of the challenge and subsequent hashing after will be set up in `config.json` and will utilize `pbkdf2` on both sides to maintain a consistent login paradigm.
```
fidian: What does "primarily up to the client" mean?  Does it mean they can hash however they want?  If so, that is probably incorrect.  Does it mean that we don't dictate?  Then that's wrong.  Perhaps it is that we don't do the hashing?  If so, then it is hashed by the client, but not "primarily" up to them.

Our goal is to tell the client how to hash the password (documentation), then to verify that it is hashed (proper length, characters), and dictate with an iron fist how the verification will be done (documentation + configuration).  We even have to dictate an example and reject things that don't match, such as when the wrong capitalization is used.  5d2e != 5D2E when you are worrying about rehashing stuff, which is what we will do with the challenge.

spec/account/account-service.spec.js
```diff
@@ -47,6 +47,7 @@ describe("AccountService", () => {
                 expect(result).toEqual({
                     accountId: "unhashedAccountId"
                 });
+                expect(storageMock.delAsync).toHaveBeenCalledWith("account/hashedContent");
```
fidian: This test should be changed to reflect that you're dealing with deleting the registration information, not the account.

lib/account/account-service.js
```diff
@@ -40,9 +40,7 @@ module.exports = function (config, secureHash, storage) {  // I would like to us
         return getDirectoryAsync(registrationDirName, regId).then(getAsync).then((accountFile) => {
             accountFile.password = accountInfo.password;
 
-            return makeAccountFileAsync(accountInfo.accountId, accountFile, options).then(() => {
-                storage.delAsync;
-            });
+            return makeAccountFileAsync(accountInfo.accountId, accountFile, options).then(storage.delAsync);
```
fidian: Oh no!  You deleted the account!

spec/storage/s3.spec.js
```diff
@@ -35,23 +34,36 @@ describe("storage/s3", () => {
...
+            expect(awsSdkMock.config.region).toBe("us-east-1");
+            s3.transit();
+            expect(awsSdkMock.S3).toHaveBeenCalledWith({
+                params: {
+                    Bucket: "test-bucket"
+                }
+            });
+
+
+            // Calling transit again to make sure we didn't call S3 again.
```
fidian: Split that into a separate test.

fidian:
Did not link [storage] to storage.md everywhere in account.md.

account-manager.js mentions @param {Object} details where the {Object} should be a typedef instead.

account-service.js uses "accountDir" elsewhere still and that refers to the account path.  They should all change.

account-manager.spec.js should expose a defaultConfig and then change create() to not take any parameters.  Should shorten tests.  I'm open to other approaches/arguments here.

lib/storage/s3.js
```diff
@@ -12,11 +12,21 @@
...
  */
 
+/**
+ * The configuration object from used to set up where the storage
+ * is located in S3 and which region to use. Only what is needed
+ * to be overridden or added has to be sent in.
+ *
+ * @typedef {Object} s3~putOptions
+ * @property {string} [contentType]
+ * @property {Date} [expires]
```
fidian: This should be OtDate

fidian:
Only 2 characters need to be added!  :-)

fidian:
Time to merge!

PR #59, opened 2016-05-03
Json Schema Validation Can Scan Recursively

AbsentSemicolon:
This was left out to be able to scan files recursively so we can organize schemas better within folders.


fidian:
Works for me.  We need to consider how this is to be initialized exactly one time in our system.

PR #62, opened 2016-05-09
Versions of Node don't work with ES6

AbsentSemicolon:
We are removing the versions of Node which don't work with ES6 so we can achieve a baseline of where the code will work.


fidian:
Yes, agreed!

PR #64, opened 2016-05-10
Login

AbsentSemicolon:
This adds the functionality for working with logging into the system.


config.json
```diff
@@ -14,6 +14,18 @@
         "initiateLifetime": {
             "hours": 1
         },
+        "loginDir": "/login/",
+        "loginHash": "sha512",
```
fidian: I would like the loginHash stuff separated from passwordHash.

```
challenge: {
    algo: "sha512"
},
passwordHash: {
    algo: "sha512",
    bytes: "48",
    ...
}
```

config.json
```diff
@@ -14,6 +14,18 @@
...
+        "loginIdLength": 24,
+        "loginLifetime": {
+            "minutes": 15
+        },
+        "passwordHash": {
```
fidian: Not totally sold on `passwordHash`.  This should be renamed, since it is pbkdf2.  It is a type of hashing, but designed to be more secure and "hash" has less meaning than "pbkdf2".

config.json
```diff
@@ -14,6 +14,18 @@
...
+            "algo": "sha512",
+            "hashLength": 48,
```
fidian: This is an ambiguous number.  Is that the length of the resulting HASH?  No.  It is the length of the resulting bytes that are generated from pbkdf2.  This number can be anything, where as a hash is a fixed length.  Maybe investigate pbkdf2 implementations in different languages and see what standard there is for the name of the length of the result.

doc/account.md
```diff
@@ -1,8 +1,7 @@
 Account
 =======
 
-Configuration
--------------
+## Configuration
```
fidian: Why did you make this particular change?

We should pick one format and consistently use it.  Previously we have used hyphens, so why switch to hash characters?

doc/account.md
```diff
@@ -12,12 +11,16 @@ The `idHash` property of `account` covers how the ids will be hashed when stored
 
 You can find a list of available hashing methods in `lib/ciphers-and-hashes.js`.
 
-Where the account information is stored can be change by updating `accountDir`. The registration files can be updated by changing `registrationDir`. Ideally these would be different as we store temporary registration files during the sign up process in the `registrationDir`, and permanent files for the account in `accountDir` once sign up is complete. These cannot be local directories as they will be passed to [storage] to handle.
-
 The lengths of the IDs we use for account creation can be adjusted using `accountIdLength` and `registrationIdLength`.
 
+There are several ways of managing how login information is created. Using `loginDir` will change where the login file is created. *This must be have a leading and ending slash.* Changing how long the id for login is changed using `loginIdLength`. Lastly how long the login file stays on the server is managed by `loginLifetime`. This should be relatively short as these login files should not stay around too long so we can fight against time attacks.
```
fidian: The "leading and ending slash" should be handled by a configuration validation module.  Would you start this module?

doc/account.md
```diff
@@ -12,12 +11,16 @@ The `idHash` property of `account` covers how the ids will be hashed when stored
...
+
+The `passwordHash` section contains how the  passwords will be hashed when creating a password and logging in. This contains the same information as `idHash` as we use the same technique to hash the passwords just with different properties. The salt will be filled in when sent to the client. **These settings should not be changed after being set up.**
```
fidian: I've been thinking about this and I want to be able to change the pbkdf2 settings at any time.  This means that the pbkdf2 settings that were used to generate the hashed password (we need to figure out a term for this) should be all encoded into the account record as well.

config.json
```diff
@@ -14,6 +14,18 @@
...
+            "iterations": 100000,
+            "salt": ""
```
fidian: This should be the length of the salt, not the salt itself.  We should generate a random salt for each account.

doc/account.md
```diff
@@ -33,31 +36,55 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
...
 
 If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
 
+## Logging In
+
+Logging into OpenToken.io is a two step process, involving sending information for the client to create a password hash and OpenToken.io validating.
+
+### Initiating a Login
+
+The client will call into the application using the hashed account id and the `accountId`. This creates and login file containing the `challengeId`, and a `salt` to use for logging in. This file also contains the `accountId`.This file is then sent to [storage].
```
fidian: Space missing between third and fourth sentences.

config.json
fidian: Oh, and add in challenge.saltLength to determine how long the challenge's seed/salt should be.

doc/account.md
```diff
@@ -33,31 +36,55 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
...
+
+Information will then be sent to the client for them to create the password for logging in. This contains the `salt` which was just created for the challenge, as well as the `hash` we are expecting them to use once they have created a password hash the same was they did for confirming signup. We also send back the data used when the password was created.
+
+### Completing a Login
+
+Once the password is created on the client side and passed back to OpenToken.io, the original password sent to use is looked up using the login file containing the `accountId`. Using the original `password` and `passwordSalt` from the file, the application then hashes it the same way using the same data the client used and then compares the hashes. If they match, we authorize the [multifactor authentication](hotp.md) token sent in. If that is valid the login is complete.
```
fidian: Reword the bit about creating a password.  They don't create a password.

They have a password.
They generate a secure hash of the password.
Finally the challenge hash is created.

Maybe those are the terms we should use?  "password", "secure hash" and "challenge hash" in order to help keep this separate in people's minds?  (People includes us too.)

lib/account/account-manager.js
```diff
@@ -5,6 +5,23 @@
  */
 
 /**
+ * Information about an account holder.
+ *
+ * @typedef {Object} accountManager~accountFile
+ * @property {string} email
+ * @property {string} mfaKey
+ * @property {string} password
```
fidian: This should get renamed.  We do *not* store the password.

lib/account/account-manager.js
```diff
@@ -5,6 +5,23 @@
...
+ * @property {string} passwordSalt
```
fidian: Just confirming, this is the Base64 encoded buffer, yes?

Comments in this file about the data stored for the properties, especially for this particular field, would be very helpful.

lib/account/account-manager.js
```diff
@@ -19,6 +36,28 @@
  * @typedef {Object} accountManager~confirmedInfo
  * @property {string} mfaKey
  * @property {string} passwordSalt
+ * @property {accountManager~hashSet} passwordHashing
```
fidian: I don't see this property used anywhere.

The name of this is "confirmedInfo".  They should be delivered the password hashing information before the registration is turned into a full account.  Confirmed = email confirmed and they are finishing signup, yes?

lib/account/account-manager.js
```diff
@@ -19,6 +36,28 @@
...
+ * @property {string} iterations
+ * @property {string} salt
+ */
+
+/**
+ * What an account holder needs to create the password hash
+ * to attempt to log in.
+ *
+ * @typedef {Object} accountManager~loginInitiateClientInfo
+ * @property {accountManager~hashSet} pbkdf
```
fidian: pbkdf2 !== pbkdf1

config.json
fidian: HOWEVER, we don't want "pbkdf2" because that's a specific implementation.  How about "secureHash"?  And we should make "pbkdf2" a value...

```
secureHash: {
    algo: sha512,
    derivation: pbkdf2,
    bytes: 48,
    iterations: 100000,
    saltLength: 64
}

File: lib/account/account-manager.js (hunk 13 above)
Comment (fidian) [inline lib/account/account-manager.js]: Just a note, this doesn't appear to match the name in the config.

File: doc/account.md [hunk 14]
```diff
@@ -33,31 +36,55 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
...
 
-Completing an Account
----------------------
+### Completing an Account
 
 To complete an account, the client will need to pass in a hashed `password`, the `regId` which we gave them previously as well as a `currentMfa` code and `previousMfa` code. This will allow us to find the registration file using the hashed `regId` and looking it up in [storage]. Getting the `mfaKey` off the account we are able to then validate the user using the MFA codes they passed in.
 
 If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
 
+## Logging In
```
fidian: I do not feel that this dives into the technical details enough.  I don't think I could write a client that understands how to log in with only this documentation.

lib/account/account-manager.js
```diff
@@ -36,6 +75,25 @@
  */
 
 /**
+ * Lifetime for how long a file should last. This can
+ * contain as many properties as needed to adjust the
+ * time.
+ *
+ * @typedef {Object} accountManager~lifeTime
+ * @property {string} Hours, Minutes, Seconds
```
fidian: The property is called "Hours," ?  I don't think this is valid.

lib/account/account-manager.js
```diff
@@ -36,6 +75,25 @@
...
+ *
+ */
+
+/**
+ * Information the Account Holder sends to complete the
+ * login process.
+ *
+ * @typedef {Object} accountManager~loginCompleteInfo
+ * @property {string} password
```
fidian: This is not the password.

lib/account/account-manager.js
```diff
@@ -46,24 +104,28 @@
...
+    accountConfig = config.account || {};
+
     /**
      * Checks to see if the values for MFA are valid and will
      * throw if they are not.
      *
      * @param {string} mfaKey
-     * @param {string} previousMfa
      * @param {string} currentMfa
+     * @param {string} [previousMfa]
```
fidian: When you make this parameter optional, how are you guaranteeing that both the current and previous have been sent during signup?

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
 
     /**
+     * Creates a password hash using the same method
+     * an account holder used to create a password to login.
+     *
+     * @param {string} password
+     * @param {string} salt
+     * @return {Promise.<string>}
+     */
+    function hashPasswordAsync(accountFile, salt) {
```
fidian: This belongs in the service.  It's not a business "feature" to make sure that the password is hashed.  It is a technical need.  Without this, the software is pointless, so that means it is in the technical realm.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+        return secureHash.hashAsync(accountFile.password + salt, accountConfig.passwordHash.secondary, true).then((hashedPassword) => {
```
fidian: secondary is not a property, from what I have seen.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+    /**
+     * Creates the object for key expiration option. We need to be
+     * specific that we are expiring a file after a given time.
+     *
+     * @param {accountManager~lifeTime} lifeTime
+     * @return {s3~putOptions}
+     * @throws {Error} if we don't pass in something or it's not an object
+     */
+    function keyExpiration(lifeTime) {
+        if (!lifeTime) {
```
fidian: Should be handled by the configuration validation module.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+     *
+     * @param {string} hashedAccountId
+     * @param {string} loginId
+     * @param {accountManager~loginCompleteInfo} loginInfo
+     * @return {Promise.<*>}
+     */
+    function loginCompleteAsync(hashedAccountId, loginId, loginInfo) {
+        return accountService.getLoginFileAsync(hashedAccountId, loginId).then((loginFile) => {
+            return accountService.getAccountFileAsync(loginFile.accountId).then((accountFile) => {
+                return loginVerifyAsync(accountFile, loginFile.salt, loginInfo.password);
```
fidian: I've been thinking.  We don't need to store the salt in our file.  The login file only needs to store the result of the challengeHash(secureHash + challengeSalt).  I think that might be the way to go.  Then we don't need to store the challenge hash information in the login file.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+
+
+    /**
+     * Need to verify the password we have on file matches
+     * what the account holder is using to log in.
+     *
+     * @param {accoutManager~accountFile} accountFile
+     * @param {string} salt
+     * @param {string} hashedPassword
+     * @return {boolean}
```
fidian: This returns a promise, not a boolean.

lib/account/account-manager.js
```diff
@@ -19,6 +36,28 @@
...
+
+/**
+ * What an account holder needs to create the password hash
+ * to attempt to log in.
+ *
+ * @typedef {Object} accountManager~loginInitiateClientInfo
+ * @property {accountManager~hashSet} pbkdf
+ * @property {accountManager~challengeSet} challenge
+ * @property {string} encoding
+ *
```
fidian: Would you add a property of `mfa` and set it to `["hotp"]`?  We don't need it but it also indicates that we require MFA for the account and what MFA is needed.  It's for the client's use *only*.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+     * @return {boolean}
+     * @throws {Error} If the password hashes do not match
+     */
+    function loginVerifyAsync(accountFile, salt, hashedPassword) {
+        return hashPasswordAsync(accountFile, salt).then((passwordHash) => {
+            if (! secureHash.compare(passwordHash, base64.encode(hashedPassword))) {
+                throw new Error("Password hashes do not match");
+            }
+
+            return true;
```
fidian: Do you need to return `true`?

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+     * @param {string} salt
+     * @param {string} hashedPassword
+     * @return {boolean}
+     * @throws {Error} If the password hashes do not match
+     */
+    function loginVerifyAsync(accountFile, salt, hashedPassword) {
+        return hashPasswordAsync(accountFile, salt).then((passwordHash) => {
+            if (! secureHash.compare(passwordHash, base64.encode(hashedPassword))) {
+                throw new Error("Password hashes do not match");
+            }
```
fidian: It seems to me that it might be easier to just drop these three lines in the function above.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+     *
+     * @param {accountManager~loginInitiateDetails} details
+     * @return {accountManager~loginInitiateClientInfo}
+     */
+    function loginInitiateDetails(details) {
+        return accountService.getAccountFileAsync(details.accountId).then((accountFile) => {
+            return random.randomIdAsync(config.account.passwordSaltLength).then((passwordSaltId) => {
+                var passwordHash;
+
+                passwordHash = accountConfig.passwordHash;
```
fidian: I don't like copying objects this way.  The next line will modify the accountConfig object.  We should make a `util` module and have `.shallowCopy()` and `.deepCopy()` methods.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+                passwordHash.salt = accountFile.passwordSalt;
+
+                return {
+                    pbkdf: passwordHash,
```
fidian: Mentioned earlier, but I think this should be `secureHash`.

lib/account/account-manager.js
```diff
@@ -83,6 +145,147 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+    function loginInitiationAsync(hashedAccountId, loginInitiateInfo) {
+        return secureHash.hashAsync(loginInitiateInfo.accountId, accountConfig.idHash).then((accountIdHash) => {
+
+            if (! secureHash.compare(accountIdHash, hashedAccountId)) {
+                throw new Error("Account hashes do not match");
+            }
+
+            return promise.props({
+                challengeId: random.randomIdAsync(accountConfig.loginIdLength),
+                salt: random.randomIdAsync(accountConfig.passwordSaltLength)
```
fidian: What is the difference between challengeId and salt?  It seems to me that they are really the same thing?

lib/account/account-manager.js
```diff
@@ -110,10 +313,10 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
 
         accountInfo = {};
         accountInfo.accountId = accountId;
-        accountInfo.password = registrationInfo.password;
+        accountInfo.password = base64.decode(registrationInfo.password);
```
fidian: I am curious why this is decoded.

lib/account/account-service.js
```diff
@@ -69,6 +69,26 @@ module.exports = function (config, secureHash, storage) {
...
+     */
+    function deleteLoginFileAsync(hashedAccountId, loginId) {
+        return secureHash.hashAsync(loginId, accountConfig.idHash).then((hashedLoginId) => {
+            var loginKey;
+
+            loginKey = makeLoginKey(hashedAccountId, hashedLoginId);
+
+            return storage.delAsync(loginKey);
+        }).then(() => {
+            return true;
```
fidian: Does it matter that this returns a boolean?  Can it just return a promise that resolves or does not resolve?

lib/account/account-service.js
```diff
@@ -128,6 +181,43 @@ module.exports = function (config, secureHash, storage) {
 
 
     /**
+     * Makes a challenge key from hashes and directory names.
+     * Errors when config settings are not in place because we
+     * do not want a directory to have undefined in the key.
```
fidian: Agreed - we don't want "undefined" in the key, though config validation should be a separate module.

lib/secure-hash.js
```diff
@@ -53,7 +53,7 @@ module.exports = function (base64, crypto, promise) {
      * @param {secureHash~config} [config]
      * @return {Promise.<string>}
      */
-    function secureHashAsync(data, config) {
+    function secureHashAsync(data, config, regularEncode) {
```
fidian: I don't like this.  Not one bit.  Think of a new developer seeing our system.  Do they want regular encode?  Do they want "non-regular" encode?  Should this encode or not?

Maybe split encoding from ID generation?

lib/secure-hash.js
```diff
@@ -71,12 +71,29 @@ module.exports = function (base64, crypto, promise) {
...
 
+
+    /**
+     * Hashes data using an algorithm passed in.
+     *
+     * @param {string} data
+     * @param {string} algorithm
+     * @return {Promise.<string>}
+     */
+    function secureHashSimpleHash(data, algorithm) {
```
fidian: Do you think a file named "secure-hash" should have "secureHashSimpleHash"?

fidian:
Reviewed.  I did not look at tests because I am wondering more about the core functionality at this time.

Also, we will need to rewrite portions of the API doc.  Would you take a stab at it and see if you can make every URL that's used into one that can get logged without divulging information?  That might mean we have to send additional data (cookies, custom headers, or body) to do things like login.

lib/account/account-manager.js
AbsentSemicolon: Good idea, we just create what to compare to straight away and then just do a compare when they log in.

lib/account/account-manager.js
AbsentSemicolon: Removed this function as it wasn't needed after some changes.

lib/account/account-manager.js
AbsentSemicolon: Agreed, after moving the hashPasswordAsync method to the service, was able to move the compare up.

lib/account/account-manager.js
AbsentSemicolon: Should have been loginId instead of challengeId. This is changed.

lib/account/account-service.js
AbsentSemicolon: Yeah makes sense, it can just send the resolve.

config.json
AbsentSemicolon: Sounds good, made this happen.

config.json
AbsentSemicolon: Split it up and added the salt length

doc/account.md
AbsentSemicolon: I started having to need more sub headings, but I can use the underline for main headings and ### for sub headings under that since ---- is the same as ##.

config.json
AbsentSemicolon: We do generate a radom salt, I was just giving a place to put the salt when sending back to the client, but I can send that in a different property.

doc/account.md
AbsentSemicolon: Fixed this.

lib/account/account-manager.js
AbsentSemicolon: Lowercased these.

lib/account/account-manager.js
AbsentSemicolon: Moved this to the service

lib/account/account-manager.js
AbsentSemicolon: Weird, I took it out, but I must have left it in the tests, so it was still working. This has been corrected.

lib/account/account-manager.js
AbsentSemicolon: No I don't...

lib/account/account-manager.js
AbsentSemicolon: This won't happen anymore as I'll return the salt no in the hashing information.

lib/account/account-manager.js
AbsentSemicolon: I've changed this to use secure hash.

lib/secure-hash.js
AbsentSemicolon: Made this into two different functions to encode into base64 and one to encode into base64 but for a URI so it doesn't have slashes and pluses.

lib/account/account-service.js
AbsentSemicolon: Will be in the config validation module.

lib/account/account-manager.js
AbsentSemicolon: In your directions at https://ide.c9.io/fidian/api-examples in the login.md on line 39 it says: "Note that we decode the base64 and save the raw buffer with the account." So I have to decode the password/hash and save that in the account file.

lib/account/account-manager.js
AbsentSemicolon: It isn't but it's what I'm expecting the client to send us. Thought this would be the simplest way for them to understand what to send us.

lib/account/account-manager.js
AbsentSemicolon: We will use JSON validation for this and everywhere else we can validate.

lib/account/account-manager.js
AbsentSemicolon: Took it out, we'll add this in the config validation modules

config.json
```diff
@@ -7,14 +7,30 @@
...
+            "algo": "sha512",
+            "saltLength": 128
+        },
         "passwordSaltLength": 128,
+        "secureHash": {
+            "algo": "sha512",
+            "bytes": 48,
+            "derivation": "pbkdf2",
+            "iterations": 100000,
+            "saltLength": 64
```
fidian: What is the difference between saltLength and bytes?  These two values are a bit confusing to me.

config.json
```diff
@@ -7,14 +7,30 @@
         },
         "idHash": {
             "algo": "sha256",
-            "hashLength": 24,
+            "bytes": 24,
             "iterations": 10000,
             "salt": ""
```
fidian: `salt` should not be defined here.

config.json
```diff
@@ -7,14 +7,30 @@
...
         },
         "initiateLifetime": {
             "hours": 1
         },
+        "loginDir": "/login/",
```
fidian: You should split the login information into a new top level property because `account` is looking very crowded.

doc/account.md
fidian: Not addressed.

doc/account.md
fidian: This was not addressed.

doc/account.md
```diff
@@ -33,31 +37,57 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
...
+        "loginDir": "/login/",
+        "loginIdLength": 24,
+        "loginLifetime": {
+            "minutes": 15
+        },
+        passwordHash: {
+            "algo": "sha512",
+            "hashLength": 48,
+            "iterations": 100000,
+            "salt": ""
```
fidian: `salt` should not exist here.

doc/account.md
```diff
@@ -33,31 +37,57 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
         "initiateLifetime": {
             "hours": 1
         },
+        "loginDir": "/login/",
+        "loginIdLength": 24,
+        "loginLifetime": {
+            "minutes": 15
+        },
+        passwordHash: {
```
fidian: This should be `"passwordHash"` (with quotes).

doc/account.md
```diff
@@ -33,31 +37,57 @@ The `passwordSaltLength` is used to adjust the length of the password salt we ge
...
 If the account validates, we create the account file from the registration file, add the `password` to it, and generate an `accountId`. The file is then saved under the `accountDir` using the hashed `accountId` as the file name. Only the `accountId` is sent back to the client.
 
+Logging In
+----------
+
+Logging into OpenToken.io is a two step process, involving sending information for the client to create a secure hash from their password and OpenToken.io validating once it's sent in.
+
+### Initiating a Login
+
+The client will call into the application using their hashed account id with the `accountId` in the data. A random `loginId` and `salt` will be created. Using the `accountId`, retrieving the file containing the client information a `secureHash` will also be dervived from the `salt` we just created and parameters which will be sent to the client to create the same `secureHash` we will save in a login file. A short expiration is made as well from the current time. This expiration should be short. This information will be saved using a hash of the `loginId` we created in a directory under the hashed `accountId`.
```
fidian: This does not read well:  retrieving the file containing the client information a secureHash will also be derived from the salt we just created

Likewise, the second half of that sentence:  derived from the salt we just created and parameters which will be sent to the client to create the same secureHash we will save in a login file.

You've repeated "short".  I don't mind stressing that the time window should be very small, but perhaps another word could be used or an explanation of why the time should be limited so severely?

lib/account/account-manager.js
```diff
@@ -5,6 +5,23 @@
...
+ *
+ * @typedef {Object} accountManager~accountFile
+ * @property {string} email
+ * @property {string} mfaKey
+ * @property {string} secureHash
+ * @property {string} passwordSalt
+ */
+
+/**
+ *
```
fidian: A better description will help me understand when this is used.  Not necessarily saying "this is returned by getThingAsync" but more along the lines of "This collection of information is sent back to the client so they are able to generate password hashes" or something.

lib/account/account-manager.js
```diff
@@ -5,6 +5,23 @@
...
+ * @typedef {Object} accountManager~challengeSet
+ * @property {string} salt
+ * @property {string} hash
```
fidian: I don't think we ever send nor receive both the salt and a hash.  I believe this is wrong, but need to chat about it.

lib/account/account-manager.js
```diff
@@ -19,6 +36,28 @@
...
+ *
+ * @typedef {Object} accountManager~hashSet
+ * @property {string} algo
+ * @property {string} bytes
+ * @property {string} iterations
+ * @property {string} saltLength
+ */
+
+/**
+ * What an account holder needs to create the password hash
```
fidian: Would you change "to" into "in order to"?  It took me three tries reading this sentence to understand that the phrase "needs to create" means "is required to have in order to create" instead of "should generate"

lib/account/account-manager.js
```diff
@@ -46,24 +104,28 @@
  * @property {string} regId Unhashed registration id sent to client
  */
 
-module.exports = function (accountService, config, hotp, otDate, promise, random) {
+module.exports = function (accountService, base64, config, hotp, otDate, promise, random, secureHash) {
```
fidian: Caution:  Starting to get long!  Should move things into services or other libraries.

lib/account/account-manager.js
```diff
@@ -46,24 +104,28 @@
...
+    var accountConfig;
+
+    accountConfig = config.account || {};
```
fidian: The validation should handle this.

lib/account/account-manager.js
fidian: How does JSON validation affect the usage of this function?  How do we enforce that the previous is required in places in our code?

I guess what I am getting at is that the manager should be the one that knows for a fact that a previous one is required or that a previous one is not required.  If split into two functions, checkMfa and checkMfaPrevious or something, then we can guarantee that it works.

I brought up this point because I could pass anything falsy - "" for instance, or null, and it will pass the validation function here but perhaps we are supposed to require the previous MFA.  If the JSON schema said that it had to be a string and we screwed around with it and changed it to a number, then I could bypass the previous MFA check by using "000000".

lib/account/account-manager.js
```diff
@@ -83,6 +145,108 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
 
 
     /**
+     * Creates the object for key expiration option. We need to be
+     * specific that we are expiring a file after a given time.
+     *
+     * @param {accountManager~lifeTime} lifeTime
+     * @return {s3~putOptions}
```
fidian: This doesn't return an object.  Just an expiration date.

lib/account/account-manager.js
```diff
@@ -83,6 +145,108 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+
+
+    /**
+     * Maps information we want from the service to return
+     * to the account holder to continue the login process
+     *
+     * @param {accountManager~loginInitiateDetails} details
+     * @return {accountManager~loginInitiateClientInfo}
+     */
+    function loginInitiateDetails(details, salt) {
```
fidian: Salt is not documented.

lib/account/account-manager.js
```diff
@@ -83,6 +145,108 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+        return accountService.getAccountFileAsync(details.accountId).then((accountFile) => {
+            return random.randomIdAsync(config.account.passwordSaltLength).then((passwordSaltId) => {
```
fidian: What is passwordSaltId?  Where is it used?

lib/account/account-manager.js
```diff
@@ -83,6 +145,108 @@ module.exports = function (accountService, config, hotp, otDate, promise, random
...
+                return {
+                    pbkdf2: accountConfig.secureHash,
+                    challenge: {
+                        salt: salt,
```
fidian: Why is the salt not generated here?  I don't understand the separation between loginInitiateDetails and loginInitiationAsync.  It doesn't appear to be a self-contained chunk of work that was sliced off into this function.

lib/account/account-manager.js
fidian: When I read "password" I expect that it is a plain text password.
When I read "passwordPbkdf2Hash" and "secureHash" I do not expect a plain text password.

I think that naming it appropriately is more understandable.

lib/account/account-service.js
```diff
@@ -121,13 +187,47 @@ module.exports = function (config, secureHash, storage) {
...
+     *
+     * @param {string} accountIdHash
+     * @param {string} challengeIdHash
+     * @return {string}
+     * @throws {Error} When config settings are missing
```
fidian: No longer seems to throw, or at least doesn't intentionally throw.

lib/config-validator.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
```
fidian: We should talk about this file.  It isn't what I had expected.

lib/secure-hash.js
```diff
@@ -66,17 +66,43 @@ module.exports = function (base64, crypto, promise) {
...
+
+    /**
+     * Hashes data using an algorithm passed in.
+     *
+     * @param {string} data
+     * @param {string} algorithm
+     * @return {Promise.<string>}
+     */
+    function createHash(data, algorithm) {
+        return crypto.createHash(algorithm).update(data).digest("base64");
```
fidian: I don't think this returns a Promise.  Why does this exist?  It's also not named with *Async at the end of the function's name.

lib/secure-hash.js
```diff
@@ -66,17 +66,43 @@ module.exports = function (base64, crypto, promise) {
...
+    }
+
     return {
-        compare: secureHashCompare,
-        hashAsync: secureHashAsync
+        compare: compare,
+        createHash: createHash,
+        secureHashEncodedAsync: secureHashEncodedAsync,
+        secureHashEncodedUriAsync: secureHashEncodedUriAsync
```
fidian: I really don't like the repetition of the module name.  `secureHash.secureHashEncodedAsync` lists "secureHash" twice.

I believe secure hashing and encoding should be separate concerns.

fidian:
I again skipped the tests.

lib/config-validator.js
AbsentSemicolon: This is removed as we are going to use JSON validation instead of this. It was supposed to be removed before updating the pull request, but I missed it.

lib/secure-hash.js
AbsentSemicolon: Yeah that wasn't the best way to name, not really sure why I did it that way.

lib/account/account-service.js
AbsentSemicolon: Taken out for we will use config validation in the near future.

lib/account/account-manager.js
AbsentSemicolon: The salt is needed to create the secure hash in the challenge file so we can compare when the client sends in their account hash. This should have been removed when hooking that up, it's left over when we were hashing after they sent in their hash for us to compare.

lib/account/account-manager.js
AbsentSemicolon: This isn't needed anymore and has been removed.

lib/account/account-manager.js
AbsentSemicolon: Removed the default.

lib/account/account-manager.js
AbsentSemicolon: Yeah that sounds better.

config.json
AbsentSemicolon: This wasn't being used actually, so I removed it.

config.json
AbsentSemicolon: Took out salt.

config.json
AbsentSemicolon: I've split out the information used exclusively for login.

doc/account.md
AbsentSemicolon: Added the pbkdf2 settings to the account file.

lib/account/account-manager.js
AbsentSemicolon: Made current and previous mfa checking separate functions.

doc/account.md
AbsentSemicolon: Took out the salt property.

doc/account.md
AbsentSemicolon: Added quotes

lib/account/account-manager.js
AbsentSemicolon: Changed to be algo in explanation and code

lib/account/account-manager.js
AbsentSemicolon: I changed it to secureHash as this matches what we are saving in the login file to compare when the client sends in their securely hashed password.

AbsentSemicolon:
I'm not sure what you mean by with, "Would you take a stab at it and see if you can make every URL that's used into one that can get logged without divulging information?" from above. Do you want a new doc which has the URLs in them and explanation, or is this a new module to do this? Either way, I'll put it on a new branch and I gather a separate pull request.

fidian:
Well, what problems can you envision if the URL is logged by every proxy in
the world?  It would log the token in the plain, which is not what we want.

--
Tyler Akins

AbsentSemicolon:
Like you said, content we don't want tracked would be tracked, like the token we send back and when a user sends one in. Likewise information the user sends like email and unhashed IDs, also information about the system, like how passwords are hashed and the length of the salts and such.

fidian:
Rejecting per our discussion.

PR #65, opened 2016-05-16
Dizzy 2 and Record Module

fidian:
Updating to use Dizzy 2, which uses a fluent interface for defining dependencies.

Added a Record module that does the serialization, double encryption, compression and handling of the "expires" information.


fidian:
Sorry, I ran the tests when I had my code but did not run the tests after I merged master.  I removed the offending line - please thoroughly check the most recent commit.

fidian:
Coverage reports appear to not be linking in automatically.
https://codecov.io/gh/opentoken-io/opentoken/pull/65

lib/buffer-serializer.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+var configured;
+
+configured = false;
+
+module.exports = (bufferSerializerModule, OtDate) => {
+    if (! configured) {
```
AbsentSemicolon: Do we do !codeHere or ! codeHere. We have both around the code done by both you and me. I'm for ! codeHere. I like the bang separated from the code/variable looks less like it's all running together.

fidian: jslint likes "!codehere", but we don't use that.

How about we don't care until we implement a JS pretty printer?

lib/dependencies.js
```diff
@@ -1,59 +1,68 @@
...
 
-var container, dizzy;
+var container, moduleDefs, Dizzy;
 
-dizzy = require("dizzy");
-container = dizzy.container();
-
-// Providing container so we can inject other modules
-// we don't want to specifically have called out here.
-container.provide("container", container);
```
AbsentSemicolon: We use container in the storage module to load what storage we are using, so currently S3. We would need this back so we can use call on the S3 module, or look at a different way to load the storage module we want to use.

AbsentSemicolon:
Reviewed and looks good.

PR #67, opened 2016-05-17
Bootstrapping and Schema Validation

fidian:
This adds JSON Schema files to validate the config.json.  Since loading
these files is async, we must now have a bootstrap phase where the
configuration is validated and the schema files are scanned and loaded
into the schema module.

This pattern can be used by other things that need to load
configuration, such as the encryption module.

The configuration is now fully tested before use.


AbsentSemicolon:
Reviewed and looks good.

PR #69, opened 2016-05-18
Updating tests to work when one directory deeper

fidian:
Moves all tests into "spec/lib" so we can make "spec/route".


fidian:
Merged master and removed the conflict.

fidian:
I've added a test for the route as well.  This brings up code coverage!

route/index.js
```diff
@@ -1,7 +1,9 @@
-module.exports = {
-    get: function (req, res, next) {
-        res.setHeader("Content-Type", "text/plain");
-        res.send("API running " + (new Date()) + "\n");
-        next();
-    }
+module.exports = () => {
```
AbsentSemicolon: Should there be a "use strict"; at the top of the file?

spec/mock/request.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
AbsentSemicolon: Should there be a "use strict"; at the top of the file?

spec/mock/response.js
```diff
@@ -0,0 +1,13 @@
+var response;
```
AbsentSemicolon: Asking again, just to make sure, should there be a "use strict"?

spec/route/index.spec.js
```diff
@@ -0,0 +1,33 @@
+describe("route: /", () => {
```
AbsentSemicolon: Last place for "use strict".

fidian:
Added "use strict" - thanks for catching that!

AbsentSemicolon:
No problem. Reviewed and looks good.

PR #70, opened 2016-05-23
Much more strict linter configuration

fidian:
This adds a much more strict linter and the necessary fixes to the codebase to allow the linter to work.  It updates the "npm run test" script to also do linting.


fidian:
This should have a d4ff commit, but I don't see it.

PR #71, opened 2016-05-23
Implement significantly stricter linter

fidian:
This adds a much more strict .eslintrc (also converts it to YAML).  I've
ignored node_modules/ and coverage/, but the rest of the files now get
scanned when running "npm test".

There were many little changes that needed to get made so the linter
would not find more problems.  It's a little strict, like jslint, but
that means our code will be more uniform.

### DIFF
--- a/debug/cipher-info.js
+++ b/debug/cipher-info.js
@@ -1,8 +1,16 @@
 #!/usr/bin/env node
+/* eslint no-process-exit:0 */
 "use strict";
 
 var algorithm, crypto, result;
 
+/**
+ * Attempt to see if a key/IV length works for a cipher
+ *
+ * @param {number} keyLen Length of key to try
+ * @param {number} ivLen Length of IV to try
+ * @throws {Error} when it does not work
+ */
 function attempt(keyLen, ivLen) {
     var iv, key;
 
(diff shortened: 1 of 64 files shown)


debug/cipher-info.js
```diff
@@ -50,21 +69,30 @@ function scan(startKey, startIv, modKey, modIv, endKey, endIv) {
...
 }
 
-function show(result, suffix) {
-    if (! result) {
-        console.log("failureFinding" + suffix + ":", true);
+
+/**
+ * Show the results
+ *
+ * @param {Object} resultObj Object returned from scan()
```
AbsentSemicolon: Should this have a typedef so we know what's contained in the object?

lib/encryption.js
```diff
@@ -110,12 +111,77 @@ module.exports = function (ciphersAndHashes, crypto, promise, random) {
...
+    }
+
+
+    /**
+     * Reads a header and generates the parameters from the data
+     *
+     * @param {Buffer} data
+     * @return {Promise.<encryption~header>}
+     */
+    function readHeaderAsync(data) {
```
AbsentSemicolon: If the functions need to be in alphabetical order, parametersFromHeader comes before readHeaderAsync.

fidian: They might be, but order is now more dictated by the necessity of defining a variable or function before it is used.

AbsentSemicolon:
Reviewed and looks good.

Issue #61, opened 2016-05-06
Make a promisifier


fidian:
With the new DI system we can make a promisifier module that takes something and promisifies it.

```
module.exports = (promise) => {
    return (otherModule) => {
        return promise.promisifyAll(otherModule);
    }
}

container.register("promisify", "./promisify").fromModule().asFactory().cached();
container.register("fs", "fs").fromModule();
container.register("fsAsync", "promisify").fromContainer().asFactory(fs).cached();
```

Filing this as an issue because it would take a little bit of time to refactor tests, but then we wouldn't need to promisify any modules inside of codebase.  We could just use factories to get the async version.

```
module.exports = (fsAsync) => {
```

fidian:
Closing this one.  Will track in #74.

PR #74, opened 2016-05-23
Promisifier

fidian:
Implemented the suggestion in #61.  Reworked the code a bit to match.  This could simplify tests because we inject the async version instead of injecting the callback-oriented one and wrapping it into an async version.


AbsentSemicolon:
Missing tests for Promisifier, brings coverage down to 99.50%

AbsentSemicolon:
Reviewed and looks good.

PR #77, opened 2016-05-30
Renaming hotp to totp

fidian:
This sorts out the naming of the MFA mechanism.

PR #78, opened 2016-05-30
Promisify node modules that are functions in themselves

fidian:
This is specifically to get the server to work again with glob.  When
you `require("glob")`, it returns a function with additional methods.
That function itself needs to be promisified as well.

### DIFF
--- a/spec/lib/promisifier.spec.js
+++ b/spec/lib/promisifier.spec.js
@@ -23,4 +23,17 @@ describe("promisifier", () => {
         expect(promisifier(obj)).toBe(result);
         expect(promiseMock.promisifyAll).toHaveBeenCalledWith(obj);
     });
+    it("promisifies a function as well", () => {
+        var fn, result;
+
+        fn = () => {};
+        result = () => {};
+        spyOn(promiseMock, "promisify").andReturn(result);
+        spyOn(promiseMock, "promisifyAll").andCallFake((x) => {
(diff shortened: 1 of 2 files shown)


fidian:
This is unrelated to #77 and may be merged in either order.

fidian:
Log time to AGI-735

spec/lib/promisifier.spec.js
```diff
@@ -23,4 +23,17 @@ describe("promisifier", () => {
         expect(promisifier(obj)).toBe(result);
         expect(promiseMock.promisifyAll).toHaveBeenCalledWith(obj);
     });
+    it("promisifies a function as well", () => {
+        var fn, result;
+
+        fn = () => {};
+        result = () => {};
+        spyOn(promiseMock, "promisify").andReturn(result);
```
AbsentSemicolon: Do you want to use spyOn here when we've been using jasmine.createSpy most other places? Do we need to specify a standard for use?

fidian: I think spyOn is appropriate here.  (We talked in person.)

fidian:
Might conflict with #80 - merge that other one first.

fidian:
Sorry, accidentally deleted this branch too.

AbsentSemicolon:
Reviewed and looks good after our chat.

PR #80, opened 2016-05-31
Lots of miscellaneous cleanups

fidian:
* Moved the use of __dirname to bootstrap.
* Changed otDate to OtDate because it is a class.
* Renamed random.randomIdAsync() to random.idAsync().
* Made all of our modules into factories.
* Renamed Storage to storage becuase it is not a class.
* Added zlib to the container.
* Inject async versions of node modules instead of wrapping them in the
  module.
* Made factories into ES6 arrow functions.
* Split secureHash.hashAsync() into secureHash.pbkdf2Async() and
  secureHash.pbkdf2IdAsync().
* Enforced config validation more through the use of schemas, which
  means I made schemas more restrictive and dropped error checking in
  the code.
* Add the container to the options passed to restify-router-magic.
* Reworked the contracts in web-server.js.
* Only use route files that do not have a leading underscore, allowing
* for libraries.
* Do not use route files with a slash because Restify does not like to
  register those.
* Added the ability to display PNG files from the API.
* Made all mock objects into factories.
* Made most mock objects automatically have spies for exposed functions.
* Made the automatic "self" link conditional, which increased the test
  complexity.


fidian:
Might conflict with #78.  Please merge this one first.  Bill time to AGI-735.

spec/lib/rest-middleware.spec.js
```diff
@@ -80,32 +80,71 @@ describe("restMiddleware", () => {
...
+            ]);
+            req.href.andReturn("/path");
+            req.method = "GET";
+            restMiddleware({
+                baseUrl: "http://localhost:8443",
+                https: false
+            }, serverMock);
+            linkFunction = serverMock.use.mostRecentCall.args[0];
+        });
+        it("set up for GET", () => {
```
AbsentSemicolon: Might be better to say "it sets up for GET" or something like that, instead of, "it set up for GET".  Just doesn't sound right.

spec/lib/rest-middleware.spec.js
```diff
@@ -80,32 +80,71 @@ describe("restMiddleware", () => {
...
+                    rel: "up",
+                    title: "self-discovery"
+                },
+                {
+                    href: "http://localhost:8443/path",
+                    rel: "self"
+                }
+            ]);
+        });
+        it("set up for POST", () => {
```
AbsentSemicolon: Same as above.

spec/lib/rest-middleware.spec.js
fidian: I see what you mean, but the subject is plural.  Look at the describe above and it ends in "links".  The "self-discovery links" are "set up for GET".

spec/lib/rest-middleware.spec.js
fidian: I don't like "sets" for the reason I detailed earlier, but I'm not opposed to changing the wording entirely.

spec/lib/rest-middleware.spec.js
AbsentSemicolon: Oh I see it now, that makes sense. I'll look at that closer going forward.

fidian:
Accidentally deleted the branch, but recovered it.  Sorry for the chaos.

AbsentSemicolon:
That's ok. Branch looks good. Even increases code coverage. Good deal.

PR #82, opened 2016-05-31
Using .json in schema files so they can be sent to clients

fidian:
Changes that are necessary to allow our schemas to be used by clients:

1.  Schema files themselves need to preserve ".json" at the end of "$ref" properties so clients can fetch them without doing fancy filename mapping.
2.  Added restify-plugins module and hook up the schema sending code.
3.  Schema library should not remove .json, which impacts code that does the validation.


fidian:
Merge this first.  Bill time to AGI-779.

AbsentSemicolon:
Reviewed and looks good.

Issue #84, opened 2016-06-02
Registration confirmation should not happen with GET


fidian:
GET implies the system state should not change.

This means we'll need to have a UI, then construct a URL to the UI with the id and confirmation code in something like the query string.  UI will then perform a POST to the right URL and confirmation will succeed.

Needs:
* UI
* Email
* Email templating
* Small amount of code to parse query strings in UI
* AJAX post to generated URL with the confirmation code in the body
* Change the route to only accept POST
* Validate incoming request against a new schema

PR #86, opened 2016-06-02
Registration is half done

fidian:
After this gets merged one is able to ...

    POST /registration
    {
        "email": "user@example.com"
    }

... and get back the beginning of a registration record.  It will have the method for how to hash the password and the MFA TOTP key.  The timer starts ticking - don't wait an hour before you complete the next step.  From here you can ...

    POST /registration/id-from-the-link-header
    {
        "passwordHash": "abcdefghijklmnopqrstuvwxyz",
        "totp": {
            "current": "000000",
            "previous": "111111"
        }
    }

... and this will (in theory) send an email with a confirmation URL.  Right now that doesn't do anything.  Assuming you could get the code by decrypting the S3 record or employing other magical spells, you can ...

    GET /registration/id-from-the-link-header/confirm/mysterious-confirmation-code

... and this will convert your record to an account as well as give you an error because it will try to generate a route for the response and there are no account-related routes.  It's not honoring the bit where GET doesn't change the server state, and issue #84 is opened to cover this deficiency.

And fetching all of the self discovery stuff and schemas work as well.  Plus now we have tests for `bin/server.js`, making those charts much prettier.  Oh, and the test bucket we are using has a policy that will clean up registrations.  And I've actually tested the record expiration code.

What's left?

* I have not confirmed that MFA works through the API with curl.  I have verified the individual components, but that doesn't guarantee things work as expected.
* Sending email or exposing the confirm link somehow.  I'd really prefer email.
    * If email, then we need email templates.
    * If email, we need to send authenticated SMTP with a module, complete with retries and logging.
    * If no email, then we need a way to confirm this stuff besides dipping into the AWS account.
* Setting up a couple account routes so Restify can at least route to them.  They do not need to work.
    * /account
    * /account/account-id/login


fidian:
Bits of information:

* Squash, don't merge on this one.  Just thought I'd call it out explicitly for this commit.  The merges were messy and they don't need to be in the git history.
* Bill time to AGI-741

lib/account/account-manager.js
```diff
@@ -1,177 +1,64 @@
...
 
     /**
-     * Validates the previous and current MFA values are valid and sends
-     * data off to the account service for updating/putting data into
-     * the account record.
+     * Create a new account record
      *
-     * @param {accountManager~registrationInfo} registrationInfo
-     * @return {Promise.<accountManager~completionInfo>}
+     * @param {accountManager~record} record
```
AbsentSemicolon: Where is the type definition for this? I don't see it in this file.

lib/account/account-manager.js
```diff
@@ -1,177 +1,64 @@
...
 
-
     /**
-     * Gets the account file from the registration id passed in
-     * so the account holder can finish signing up.
+     * Returns the password hashing settings and a new salt.
      *
-     * @param {string} regId
-     * @return {Promise.<accountManager~confirmedInfo>}
+     * @return {accountManager~passwordHash}
```
AbsentSemicolon: We talked about this one.

lib/account/account-service.js
```diff
@@ -3,154 +3,80 @@
...
  *
- * @typedef {Object} accountService~initiationFile
- * @property {string} email
- * @property {string} mfaKey
- * @property {string} passwordSalt
+ * @param {Object} config
+ * @param {Object} OtDate Helper object and Object class, like Date
+ * @param {Object} record
+ * @param {Object} storage
+ * @return {Function} factory
```
AbsentSemicolon: I see an explanation of this module,AccountService, but this information is missing at the top of AccountManager, there's no explanation and parameter breakout for the manager.

route/registration/_id/_get.js
```diff
@@ -0,0 +1,20 @@
...
+            title: "registration-secure"
+        },
+        item: {
+            href: server.router.render("registration-secure-qr", {
+                id: secureInfoGroup.id
+            }),
+            title: "registration-secure-qr"
+        }
+    });
+    res.send(secureInfoGroup.secureInfo);
```
AbsentSemicolon: Should we have a code here, or is it an implied 200?

route/registration/_id/qr/index.js
```diff
@@ -0,0 +1,23 @@
...
+                    res.contentType = "image/png";
+                    res.links({
+                        item: {
+                            href: server.router.render("registration-secure", {
+                                id: req.params.id
+                            }),
+                            title: "registration-secure"
+                        }
+                    });
+                    res.send(pngData);
```
AbsentSemicolon: Same as above, is this an implied 200?

AbsentSemicolon:
From the comments we talked about about this looks good to me.

PR #87, opened 2016-06-23
This is already cached in dependencies.js

fidian:
By making this change, branch-related code coverage increases.


fidian:
No dependencies on other pull requests.  AGI-735.

PR #91, opened 2016-06-23
Changing name and removing an argument

fidian:
I can save a bit of work with each invocation by converting the "GET"
response handling function into a factory.


fidian:
No dependencies on other pull requests. AGI-735.

AbsentSemicolon:
Reviewed and looks good.

PR #92, opened 2016-06-24
Consolidating services for the managers

fidian:
They did the same basic operations - set, get, delete.  For the upcoming
changes and additional managers that will be committted, they too only
used these operations.  I've made the services into a single service
factory.


fidian:
AGI-640, no dependencies on other pull requests.

AbsentSemicolon:
Reviewed and looks good.

AbsentSemicolon:
Reviewed and looks good.

PR #94, opened 2016-06-24
Rename secureHash and split off encoding

fidian:
Renames secureHash to just hash in preparation for future work.  The
secureHash.hashAsync method is renamed to hash.deriveAsync because it is
a key derivation function.  In the future there will be a simpler
version that just does a single has and it will be hash.hashAsync.

Split off the encoding to make it more modular, key it off config, allow
it to use other modules and to allow other modules to use it directly.


fidian:
AGI-640, no dependencies because you merged them all in already.

AbsentSemicolon:
Reviewed and looks good.

PR #95, opened 2016-06-27
Avoid shadowing of variable in scope

fidian:
`promise` is already defined.  Sadly, this passed tests but would not
work in real life as it redefines `promise` in the outer scope.


AbsentSemicolon:
Reviewed and looks good.

PR #98, opened 2016-07-20
Allowing multi-part IDs for storage

fidian:
You enable them by passing multiple ID hash config objects to the
storage service factory.  The storage service will then also receive an
array for the `id` of the record to manipulate.  Each part of the array
will map to the hash at the same position.  If there's more IDs than
hashes defined, the hash definitions will loop.

This also removes some unnecessary dependencies in a couple modules and
consolidates code for duplicating objects, which may help prevent
problems with objects changing in memory (though that's not supposed to
happen in the first place).


fidian:
I brought up the unit tests back to 100%.  Sorry about missing that line.

lib/util.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = () => {
+    /**
+     * Return a deep copy of an object.
+     *
+     * @param {*} input
+     * @return {*} copy
+     */
+    function clone(input) {
+        if (typeof input === 'object' && input) {
```
AbsentSemicolon: Like Travis-CI found. Need to be double quotes.

lib/util.js
```diff
@@ -0,0 +1,22 @@
...
+        }
+
+        return input;
+    }
+
+
+    return {
+        clone
+    };
+}
```
AbsentSemicolon: Like Travis-CI found. Missing a semicolon.

AbsentSemicolon:
Reviewed and looks good.

spec/lib/util.spec.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+describe("util", () => {
+    var util;
+
+    beforeEach(() => {
+        util = require("../../lib/util")();
+    });
+    describe(".clone()", () => {
+        it("makes a deep copy", () => {
```
AbsentSemicolon: Need another test which doesn't pass in an object. We are only testing one branch of utils.

AbsentSemicolon:
Reviewed and looks good.

PR #99, opened 2016-07-21
Using the storage selection file

fidian:
We were using the S3 module directly, though that is not what we should
have been doing.  The purpose of storage.js is to look at the config and
select the correct storage engine to use.


AbsentSemicolon:
Reviewed and looks good.

PR #100, opened 2016-07-21
Remove session on logout, route changes, other cleanups

fidian:
On logout the session record is now removed from storage.

The router is updated to allow _account-id to be :accountId.  This was
updated so we can now have /account/_account-id/something/_something-id
in order to keep the IDs separate.

Other things were found and updated as well.


lib/account-manager.js
```diff
@@ -135,6 +135,18 @@ module.exports = (challengeManager, config, random, sessionManager, storageServi
 
 
     /**
+     * Destroy the session record.
+     *
+     * @param {string} accountId Account ID
```
AbsentSemicolon: Do we need a comment saying Account ID when the parameter is `accountId`? I feel this is self-explanatory. Same with Session ID below this.

lib/session-manager.js
```diff
@@ -25,6 +25,21 @@ module.exports = (config, promise, random, storageServiceFactory) => {
 
 
     /**
+     * Destroys a session
```
AbsentSemicolon: Missing period.

lib/account-manager.js
fidian: I completely agree.  Commit 7274227

lib/session-manager.js
fidian: Arrrrggggg.  9a499d0

AbsentSemicolon:
Reviewed and looks good.

PR #101, opened 2016-07-28
Style guide

fidian:
Adding style guides

Would you four want to peruse them?
@kyle-long @AbsentSemicolon @lnagle93 @quantumew


doc/style-guide-javascript.md
```diff
@@ -0,0 +1,185 @@
...
+* Filenames
+    * Tests for `*.js` files are named `*.spec.js`.
+    * Tests live either with the JavaScript being developed or within a `spec/` folder.  If the folder exists, all tests must be within it.
+    * Angular directives, services, etc. have their type after the name (eg `some-directive-service.js` and `datepicker-directive.js`).
+    * Related files should be named similarly (eg `response-template-service.js`, `response-template-service.css`, `response-template-service-declined.html`, `response-template-service-approved.html`).
+* Capitalization and Naming
+    * Variables and methods should use `camelCase` style formatting.
+    * Only constructors should use `UpperCamelCase` formatting, such as `Date` and `Promise`.
+    * No underscore before, within, nor after variables, functions, or properties.
+    * Functions shall be named (eg. `function MyFunc()`) only if they are not ever supposed to be overridden.  Anonymous functions don't need a name.
```
quantumew: I am a little confused here. Are you saying that functions that are not ever supposed to be overridden should be `UpperCamelCase`? The wording on this particular line threw me off a little bit but I am probably just slow.

doc/style-guide-javascript.md
```diff
@@ -0,0 +1,185 @@
...
+    * There shouldn't be any other newlines.
+    * Spaces before `{`, `[`, `(` unless preceded by another one of those symbols.  Same goes for a space after the closing `}`, `]` and `)`.  Do not add a space before `(` when calling a function or defining a named function.
+    * ESLint forces `case` statements and the `default` case in switches to be un-indented (just that line).  There must be a blank line after `break;` as well.
+* Inline Objects and Arrays, Defining Objects
+    * No trailing comma after the last property or element.
+    * Property names shall be in alphabetical order, followed immediately by a colon, a space, and the property's value.
+    * Property names should be unquoted unless it is necessary to wrap it in quotes.
+    * When defining a class, its methods are all alphabetized.
+* Order of items within a function
+    * ESLint flags (only if necessary)
```
quantumew: Missing period here.

fidian: 5b30ee8

doc/style-guide-javascript.md
fidian: Yeah, I fixed this wording.
5b30ee8

doc/style-guide-programming.md
```diff
@@ -0,0 +1,24 @@
...
+    * Files contain one class or one feature.
+    * For hyphenation purposes, acronyms and initialisms are treated as single words.  Your file that parses HTML may be called `html-parser.c`.
+* Indentation and Whitespace (mostly handled with [EditorConfig](http://editorconfig.org/)
+    * 4 spaces per each level.
+    * No spaces at the end of a line.
+    * No tabs anywhere.  If you wish to use a tab, use `"\t"` instead.
+    * The last line in a file shall have a newline after it.
+* Ordering
+    * Properties in objects will be ordered alphabetically.
+    * Class methods will be ordered alphabetically.
```
kyle-long: What if these methods have dependent methods?  Do the dependent methods come before regardless of name?

doc/style-guide-programming.md
```diff
@@ -0,0 +1,24 @@
...
+    * When there are no other dependencies on other functions, individual functions in a file will be ordered alphabetically.  When your functions in a module call each other, the dependent ones must be placed earlier.
```
kyle-long: After ordering by dependency, it is ordered alphabetically I assume?

doc/style-guide-programming.md
```diff
@@ -0,0 +1,24 @@
...
+    * When using dependency injection, the parameters will be alphabetized.
```
kyle-long: Is this a hard and fast rule?  For instance, you have a set of dependencies composing something.  All of them take a `container` which typically comes first but in one case a class needs to take an `aardvark`.

doc/style-guide-programming.md
fidian: Yep.  I was just calling out the part that some functions must be placed earlier.

doc/style-guide-programming.md
fidian: The container shouldn't be passed in.  If you were thinking of `config`, then I don't see a reason that `aardvark` can't come before `config` in the factory.  If you're talking about something special, such as the `self` parameter in Python, then I would give preferential treatment to that argument.  I think alphabetization should come second when there is a more pressing reason to have variables in a particular order.

kyle-long: Yes, that was the meat of the question, however, container may be passed in other dependency injection systems.  In fact that was the original point of the container

kyle-long: Also can we add something to that effect?  It seems like an important exception.

doc/style-guide-css.md
```diff
@@ -0,0 +1,62 @@
...
+    * Class names use hyphens, not camel case.
+    * Rules are alphabetized by the class name.  Alphabetize numbers (including negative numbers) in a "human sort" (by their value) so "neg2" comes before "1".
+    * Be *very careful* with specificity.  We aim to have rules with very low specificities.
+* Namespacing
+    * Typically, atomic rules will not require any namespacing beyond "body".
+    * If you have special CSS for a particular template make sure to namespace it with the following rules:
+        * In your template (eg. "template.html") give the top level element a class that matches your template (eg. `<div class="template">`).
+        * Create a new file ("template.less") that will nest all of the special rules under the `.template` class.
+
+    /* This is a sample of some valid LESS.
```
kyle-long: This next block doesn't appear to show up as a code block for me.

doc/style-guide-git.md
```diff
@@ -0,0 +1,21 @@
+Style Guide for Git
+===================
+
+This extends the [general programming](style-guide-programming.md) portion of the [Style Guide](style-guide.md).
+
+
+* Branches
```
kyle-long: Many times I've had problems with people not cleaning up old branches.  Can we have a note about that or add to the last note as part of "maintain the branch"

doc/style-guide-html.md
```diff
@@ -0,0 +1,39 @@
...
+* Elements
+    * XHTML style elements: `<span class="tt-u">Test</span>`, `<br />`.
+    * Attributes have spaces before them.
+    * No spaces on either side of `=` nor inside the value unless necessary.
+* Indentation
+    * Block and inline-block elements will always have a newline and indent the content within.
+    * Newline after `<br />`
+    * No other newlines.
+
+    <div>
```
kyle-long: I think this is supposed to also show up as a code block.  It is not for me.

doc/style-guide-javascript.md
```diff
@@ -0,0 +1,185 @@
...
+    * Your code.
+* Specific Constructs
+    * Nothing that's displayable to the user should be sourced from JavaScript.  It should come from an HTML template, a JSON file from a server or API calls.
+    * Use double-quoted strings.
+    * We code for ECMA Script 6.
+    * All control structures, like `if` statements, require braces.
+    * Ternary statements should be avoided.
+    * Avoid manual minification - let a minifier do it.
+        * Ternary statements should be expanded into `if` blocks.
+        * While `!!~haystack.indexOf(needle)` is tricky and neat, write `str.indexOf(needle) !== -1` so all programmers on the project immediatly know what the code does.
```
kyle-long: Getting pretty nit picky but in one example you use `haystack` and in the next you use `str`

doc/style-guide-html.md
```diff
@@ -0,0 +1,39 @@
...
+        This is a very long comment which explains how we are iterating over the
+        child nodes.  It is wrapped at 80 characters and the tags are on separate
+        lines.
+    -->
+    <div class="ta-c z-5 ml-12" ng-repeat="childNode in node.childNodeList index by idx" my-custom-directive="childNode">
+        <div>
+            {{childNode.name}}
+        </div>
+        <p>
+            You will enjoy the <em>{{childNode.manufacturer}} {{childNode.model}}</em>!  {{childNode.description}}
```
AbsentSemicolon: We should be wrapping angular text in spans or another tag as using `{{childNode.manufacturer}}` will show in the html before angular runs and can be unsightly. If you do `<span ng-bind="childNode.manufacturer"></span>`, it will not show until angular runs and the scope is updated.

doc/style-guide-javascript.md
```diff
@@ -0,0 +1,185 @@
...
+        * `undefined` is an empty value that usually signifies a missing return value, parameter or unsent data.  If we assign it to something it should be considered some sort of error.
+        * Most conditional logic should be fine using `if (cond)` instead of `if (cond === true)`.  Avoid `if (!cond === false)`.
+    * JavaScript that is expected to work outside of Angular is written using a module loading syntax.  Angular-specific code uses a module syntax.
+* Testing
+    * Tests should be written for all JavaScript code to ensure it works as expected.  100% test coverage, including testing all branches, is possible.
+    * We wish to perform behavioral testing.  Test every intended way we will be using objects and all exposed functions.
+    * All tests should pass before merging code to a parent branch.
+    * Tests should be grammatically correct and read like a sentence.  The descriptions should describe the observable behavior, not what is going on in the code.
+
+    /**
```
kyle-long: Again, this is not showing up as a code block.  Is it just me?

doc/style-guide-markdown.md
```diff
@@ -0,0 +1,42 @@
+Style Guide for Markdown
+========================
+
+This extends the [general programming](style-guide-programming.md) portion of the [Style Guide](style-guide.md).
+
+
+* Markdown Formatting
+    * Follow GitHub Flavored Markup.
+    * Paragraphs should be all on one long line.
```
kyle-long: What if its a long paragraph?  Do you want 300 characters on the same line?

doc/style-guide-markdown.md
```diff
@@ -0,0 +1,42 @@
...
+    * Avoid HTML markup.
+    * Prefer to use indentation instead of code fences.
```
kyle-long: I think we chatted about this at one point but I forgot the reason for it.  Readability when editing?  I ask because I find indentation doesn't work as well as code fences especially for other gitlab (must have a space at the end of a line if in a list?).

doc/style-guide-javascript.md
```diff
@@ -0,0 +1,185 @@
...
+            this.callCount += 1;
+        };
+
+        // Export to UMD or the global object - standard boilerplate
+        if (module && module.exports) {
+            module.exports = FakeTestClass;
+        } else {
+            global.FakeTestClass = FakeTestClass;
+        }
+    }(this, window))
```
AbsentSemicolon: I think we need to define how to use `"use strict";`. I like the function form of it when dealing with Angular since we should wrap everything in `(function () {})();` anyway and this keeps use strict within the function scope when we concatinate all files. But for Node files it's probably more appropriate to put it at the top of the file since those files won't get minified into one like we do with Angular files

doc/style-guide-markdown.md
```diff
@@ -0,0 +1,42 @@
...
+* Headings and Subheadings
+    * Documents start with a heading.  No other place uses a first level heading.
+    * Prefer `====` and `----` for first and second level headings.
```
AbsentSemicolon: Do you want the `====` to be as long as the content above it?

Something Here
============

doc/style-guide-markdown.md
```diff
@@ -0,0 +1,42 @@
...
+* Whitespace
+    * Indent 4 spaces whenever there's indentation.
+    * Two blank lines before any headings unless they are at the top of a file.
+    * One blank line after headings.
+    * One blank line before and after lists.
+* Attention to Detail
+    * Spell check and attempt to grammar check your work.
+
+    My Sample Document
```
kyle-long: This is showing up as normal markdown for me

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+
+Guidelines
+----------
+
+The [style guide](style-guide.md) lists many useful tips and various requirements when writing tests.  Only a few of those are listed here for further explanation.
+
+
+### Eliminate Duplicated Setup
+
+When the code has two separate actions that it performs then you want to have the tests split into two separate `it()` calls.  These would require the same setup to achieve an identical scenario.  Move that into a `beforeEach()`.  When you have a complicated piece of code that has numerous logical branches, again you will want to split the setup into multiple `beforeEach()` calls, one for each branch.
```
AbsentSemicolon: Should we mention `beforeEach` is for Jasmine? What if they want to use a different testing framework.

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+
+There's examples in this document to help explain this process.
+
+
+### Make it easy to expand
+
+Let's say that your code currently does only one thing.  Perhaps it just trims whitespace.  You'd see no reason to have a `beforeEach()` that would set up some service and then a separate `it()` to test if the whitespace was removed.  You could put all of the setup and the testing together in an `it()`.
```
AbsentSemicolon: Like the `beforeEach`, `it()` might be specific to Jasmine.

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+                        $rootScope.$apply();
+                    });
+                    it('sets only the "even" property', function () {
+                        expect($scope.even).toBe(true);
+                        expect($scope.odd).toBe(undefined);
+                        expect($scope.number).toBe(undefined);
+                        expect($scope.error).toBe(undefined);
+                    });
+                });
+                describe('when fulfilled with an odd number', function () {
```
kyle-long: Style guide says to prefere double quotes.  I get some of the other describes and its since they include double quotes in the string.

Is it more important to be consistent throughout a file?

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+
+It is recommended to not do this.  The `it()` is considered the "test" portion of the code and `beforeEach()` is reserved for setup.  Think about what you would have to do in order to prove a bug was fixed properly.  If the setup was combined with the test, then you'd either have to duplicate the setup in the second test (already you know that's bad) or you'll have to refactor the test to have a `beforeEach()` and two `it()` calls.  Since the latter option is the one we'd like to see, try to make all setup happen in a `beforeEach()` even if there is only one test.  It makes the tests easier to expand.
+
+
+### Test Behavior
+
+We want to write tests to not only confirm there are no defects, but also ones that could mimic how the software is intended to be used.  By testing behavior, we are shifting our viewpoint to be from the user's point of view instead of being overly concerned with the internals of the software.  In essence, I want to make sure your code performs a specific action to produce the desired result.  I do not want to guarantee that the code does every specific step in exactly the right order.
```
AbsentSemicolon: I think we should stay with the `we` context. Switching to `I` in the middle seems a bit odd. 

    In essence, we want to make sure the code performs a specific action to produce the desired result.  We do not want to guarantee the code does every specific step in exactly the right order.

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+            };
+        }));
+        describe('initial compile', function () {
+            var allowNoValueEvent;
+
+            // By default we do not want the "no value" event emitted
+            beforeEach(function () {
+                allowNoValueEvent = false;
+                $parent.on('no value', function () {
+                    expect(allowNoValueEvent).toBe(true);
```
kyle-long: Was told to flag after some discussion.  Doing an assertion in a `beforeEach`

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+-----------
+
+This is a sample test for a directive in Angular.  Let's pretend our directive will simply set some property on an isolated scope to the value passed into the directive.  There are many things wrong with this test.
+
+    /*global angular, describe, expect, inject, it, module*/
+    describe('myCustomDirective', function () {
+        var $scope;
+
+        beforeEach(module('app'));
+        it('some property on scope', inject(function ($compile, $rootScope) {
```
AbsentSemicolon: You're using single quotes here when the style guide for JavaScript says to use `double` quotes.

doc/style-guide-html.md
fidian: I do understand what you are saying, but that is not possible when the HTML is really a template.  The flicker you're talking about is only during application load.

Perhaps I should add a comment in the text indicating that this is an Angular template?

doc/style-guide-markdown.md
fidian: Yes.  That's why you've been seeing the style guide paragraphs as one line.

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+                ]
+            });
+        });
+    });
+
+What is wrong?
+
+* The data passed in as $scope.test does not need to have a lot of data.  What the directive accomplishes is merely setting the directive's `$scope.some` to the value itself.
+
+* Since we know the object that `$scope.some` should be, we should use `.toBe()` instead of `.toEqual()`.
```
AbsentSemicolon: I don't think you can use `.toBe()` when comparing objects since the instance of those objects are different, hence when comparing objects you need to use `.toEqual()`.  That's they way I've always had to do it when using Jasmine.

doc/style-guide-markdown.md
fidian: Indentation in lists is tricky because you have to double-indent the code.  Without the newlines before and after, some parsers (like GitHub) do not recognize the code block.  Here's an example of code inside of a list.  Note that I used "1." for both of the numbers in the comment when you view the raw version but the markdown-parsed version shows "2." for the second item.

1. list item

        code

1. another list item

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+
+That sure is a lengthy example, but carefully notice how anyone could add additional assertions, test cases, alternate setups and different scenarios.  The event testing code is clearly separate from the initialization code.  Promises are used for resolving values.  Events are sent in the correct way through the directive and are tested at the appropriate levels.  Data is flowing through the directive and still you don't tie yourself to any other services or code outside of your directive.
+
+
+Data Sets
+---------
+
+A filter is often tested by providing a series of inputs and asserting the output is correct.
+
+```
```
AbsentSemicolon: Why are you using a code block here when you don't other places and specifically say in your guid for markdown to not use code blocks but to indent?

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+============
+
+Sometimes a directive may call an internal function that performs specific actions.  In these cases you want to ensure those things do exactly what they should do if that internal function is called.
+
+    /*global angular, beforeEach, describe, expect, inject, it*/
+    describe('myDirective', function () {
+        var compile, eventData, eventList;
+
+        function testEffectsOfInternalFunction(scope) {
+            it('emits "goAhead", function () {
```
AbsentSemicolon: Missing a single quote after `"goAhead"`.

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+
+This may seem to contradict the "Test Behavior" section because we must be able to inspect the code to see every possible logical path that would be taken.  There is no conflict - each logical path results in outward behavior that the user might encounter.  Even if the code is later refactored to eliminate many logical paths, the tests should still pass and there would be no need to delete any of the tests that now follow a redundant path.
+
+
+Sample Test
+-----------
+
+This is a sample test for a directive in Angular.  Let's pretend our directive will simply set some property on an isolated scope to the value passed into the directive.  There are many things wrong with this test.
+
+    /*global angular, describe, expect, inject, it, module*/
```
kyle-long: This whole example should use arrow functions shouldn't it?

doc/style-guide-testing.md
```diff
@@ -0,0 +1,467 @@
...
+                var scope;
+
+                scope = $rootScope.$new();
+                $compile(angular.element('<div my-directive="input"></div>'))(scope)
+
+                // Directive makes its own isolated scope
+                return scope.$$childHead;
+            };
+        }));
+        it('allows numbers', function () {
```
kyle-long: For the few tests below it looks like they do the same thing but the descriptions in `it()` seem to suggest it should be testing something different.

doc/style-guide-programming.md
kyle-long: I see

doc/style-guide-programming.md
kyle-long: got ya

doc/style-guide-html.md
AbsentSemicolon: I'm having trouble seeing places I think are supposed to be code blocks.

doc/style-guide-javascript.md
AbsentSemicolon: This does not show up as code block.

doc/style-guide-markdown.md
kyle-long: This doesn't really answer my question.  I often have similar issues (which I can start to document if you'd like) with indents instead of code fences.  Why do you prefer indentation?

doc/style-guide-javascript.md
kyle-long: I think if you are going to concatenate a lot of javascript files together it should be function level, otherwise file level.  I don't believe this has anything to do with minification.  Something I am missing?

doc/style-guide-testing.md
kyle-long: hah, beat you to it

doc/style-guide-html.md
AbsentSemicolon: That would probably be a good thing to note for people unfamiliar with Angular.

doc/style-guide-javascript.md
AbsentSemicolon: My main point is that we don't have anything about use strict in the guide. I believe it should be in the style guide just to make sure people who reference this, who might be unfamiliar with JS would know we require it and then there could be a bit about if you minify use the function form. Otherwise at the top of the file under the globals definition, if globals is required.

doc/style-guide-testing.md
AbsentSemicolon: Ha, you did, just saw your comment.

fidian:
Will move this to its own site.

fidian:
See https://tests-always-included.github.io/style-guide/

PR #102, opened 2016-08-05
Access code

fidian:
One can now register, login, and create access keys.  Those keys can not be deleted yet.

Testing was tedious, so I added example scripts to help you communicate with an API.  This exposed some problems so I also fixed those problems.

Created a helper to make testing of links easier.  Updated a lot of tests to use this helper.

Created another helper to wrap middleware so it acts like a promise. This took code out of the route tester helper and allows anyone to use it wherever they like, such as in tests.

Moved middleware to `lib/middleware/`, including the `_login-cookie.js` middleware from the `route/` folder.

Created more middleware to validate a session cookie.

Also a ton of little other things.


fidian:
Log time to AGI-739

example/create-api-keys.js
```diff
@@ -0,0 +1,124 @@
...
+    } else {
+        description = "";
+    }
+
+    if (!config.password || !config.api || !config.mfa || !config.account) {
+        console.log("You are missing a required setting.  This needs");
+        console.log(" * API URL");
+        console.log(" * Account ID");
+        console.log(" * Account password");
+        console.log(" * MFA key (the TOTP seed, base32 encoded)");
```
AbsentSemicolon: Would it be good to match other of the doc for the options? Just to avoid some confusion.

example/create-account.js
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env node
+"use strict";
+
+var container, exampleLib, hash, neodoc;
+
+/**
+ * Returns the usage instructions
```
AbsentSemicolon: Needs a period.

.istanbul.yml
```diff
@@ -1,5 +1,6 @@
 instrumentation:
     excludes:
+        - example/**/*.js
```
AbsentSemicolon: Do you care these are alphabetical, or going longer to shorter?

lib/access-code-manager.js
```diff
@@ -0,0 +1,69 @@
+"use strict";
+
+module.exports = (config, promise, random, storageServiceFactory) => {
+    var codeLength, secretLength, storageService;
+
+    /**
+     * Create a set of API credentials
```
AbsentSemicolon: Needs a period.

lib/access-code-manager.js
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @param {string} accountId
+     * @param {accessCodeManager~createRequest} request
```
AbsentSemicolon: Are you missing the type def for this. I would thin it'd be in this file since this is the Access Code Manager.

lib/access-code-manager.js
```diff
@@ -0,0 +1,69 @@
...
+     * @return {Promise.<accessCodeManager~codeInfo>}
```
AbsentSemicolon: Same as above, seem to be missing the type def.

example/create-api-keys.js
fidian: I put them in this order because that's how they are used.  I put the options alphabetically so they are easy to reference.  I can't say that this list makes more sense.

* Account ID
* API URL
* MFA key (the TOTP seed, base32 encoded)
* Account password

I would imagine I always want the account password next to the the account ID.  Thoughts?

example/create-account.js
fidian: Fixed.

.istanbul.yml
fidian: Well, if there are exclusion patterns then they might require a special order.  Otherwise I'd lean towards alphabetical.  Fixed.

lib/access-code-manager.js
fidian: Fixed.

lib/access-code-manager.js
fidian: Great catch!  I remember distinctly thinking that I need to define this as I typed the jsdoc, but I didn't make a note of it so I promptly forgot.

lib/access-code-manager.js
fidian: Agreed.  Fixed.

fidian:
Most of the things have been fixed.  I have an open question back to you, @AbsentSemicolon, regarding the order of items that are cited in a help message.

example/create-api-keys.js
AbsentSemicolon: Ah, that makes sense since this would be the order the user would see them, so that would be beneficial to the user .

lib/storage-service-factory.js
```diff
@@ -6,6 +6,11 @@
  * for the hash that's used for the storage service.
  */
 
+/**
+ * @typedef {Object} storageServiceFactory~recordOptions
+ * @prop {OtDate} expires Expiration date of the record
```
AbsentSemicolon: This should be property.

AbsentSemicolon:
Reviewed and looks good.

PR #103, opened 2016-08-25
Updating the validation middleware

fidian:
This renames the middleware to indicate that it only validates the body.
There will be a new middleware that validates the query string, which is
why this change needed to be made.

I also extracted the mock middleware into its own mock file and had to
then update four route tests to inject the mock into the container.

The majority of the rest of the changes are simply updates to use the
new middleware's name.

### DIFF
--- a/spec/mock/middleware/validate-request-body-middleware-mock.js
+++ b/spec/mock/middleware/validate-request-body-middleware-mock.js
@@ -7,7 +7,7 @@ container = require("../../../lib/dependencies");
 module.exports = () => {
     var middleware, middlewareFactory;
 
-    middleware = jasmine.createSpy("validate-request-body-middleware-mock-middleware").andCallFake((req, res, next) => {
+    middleware = jasmine.createSpy("validateRequestBodyMiddlewareMockMiddleware").andCallFake((req, res, next) => {
         var result;
 
         result = container.resolve("schema").validate(req.body, middleware.schemaPath);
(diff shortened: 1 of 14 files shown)


fidian:
The first of hopefully many.
AGI-642
@AbsentSemicolon and @lnagle93 please review.

spec/mock/middleware/validate-request-body-middleware-mock.js
```diff
@@ -0,0 +1,32 @@
+"use strict";
+
+var container;
+
+container = require("../../../lib/dependencies");
+
+module.exports = () => {
+    var middleware, middlewareFactory;
+
+    middleware = jasmine.createSpy("validate-request-body-middleware-mock-middleware").andCallFake((req, res, next) => {
```
AbsentSemicolon: Other places you create a spy you do camel casing, we should match here as well.

spec/mock/middleware/validate-request-body-middleware-mock.js
```diff
@@ -0,0 +1,32 @@
...
+            res.send(400, new Error(`Did not validate against schema: ${middleware.schemaPath}`));
+
+            return next(false);
+        }
+
+        return next();
+    });
+
+    // This factory only returns the same middleware over and over.
+    middlewareFactory = jasmine.createSpy("validate-request-body-middleware-mock-factory").andCallFake((schemaPath) => {
```
AbsentSemicolon: Other places you create a spy you do camel casing, we should match here as well.

AbsentSemicolon:
Done with first review.

fidian:
Both are addressed in commit 360546b

AbsentSemicolon:
Reviewed and looks good.

lnagle93:
Late to the party, but I didn't see anything that needed changing anyway  ¯\_(ツ)_/¯

PR #104, opened 2016-08-25
Better error handling with error objects and formatters

fidian:
The errorResponse module now can make ErrorResponse objects.  If you
pass an ErrorResponse object to `res.send()` it will be formatted as a
standard error output.  To do that, all of the formatters needed to be
rewritten because I could not inject middleware or a prefilter of any
sort into Restify.

This also breaks all of the other formatters out into separate files,
making testers much easier.

Restify also (weirdly) never sets the Content-Length header and relies
on the formatters to do that.  Humbug, I say.  So I decided to
consolidate that functionality in the generic formatter.

This also had a lot of repeat code, so I made yet another helper to make
testing formatters easier.


fidian:
AGI-642
@AbsentSemicolon and @lnagle93 please review.

lib/formatter/error-json-formatter.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+module.exports = (errorResponse, genericFormatter) => {
+    // The default function here should not be called because this content
+    // type should be handled by genericFormatter.  So, if execution does
+    // come to this function then I don't know what `body` really is.
+    return genericFormatter.formatWithFallback((req, res, body, done) => {
+        errorResponse.createAsync("Unknown error", "4qsTYJa3").then((errorObject) => {
```
AbsentSemicolon: Is `4qsTYJa3` supposed to be random, or do we always want this code?

lib/formatter/json-formatter.js
```diff
@@ -0,0 +1,17 @@
+"use strict";
+
+module.exports = (genericFormatter) => {
+    return genericFormatter.formatWithFallback((req, res, body, done) => {
+        var string;
```
AbsentSemicolon: Not sure I like the idea of using a variable named a type, maybe use `bufferContent`, or something else, just don't like `string`.

spec/lib/error-response.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        randomMock = require("../mock/random-mock")();
+        errorResponse = require("../../lib/error-response")(configMock, loggerMock, randomMock);
+    });
+    it("exports known functions", () => {
+        expect(Object.keys(errorResponse).sort()).toEqual([
+            "createAsync",
+            "rejectedPromiseAsync"
+        ]);
+    });
+    describe("ErrorResponse objet", () => {
```
AbsentSemicolon: `objet` should be `object`

lib/formatter/error-json-formatter.js
fidian: Each error is given a random code in our codebase.  Every time that error is produced it will have the same garbage code as an identifier.  Developers using the API should be able to trust that these codes don't change unless the meaning of the error also changes.

lib/error-response.js
```diff
@@ -0,0 +1,75 @@
...
+         * @return {string}
+         */
+        mimeType() {
+            return "application/vnd.error+json";
+        }
+    }
+
+
+    /**
+     * Creates an error object
```
fidian: 3e62e73

spec/lib/error-response.spec.js
fidian: 3acc9bd

lib/formatter/json-formatter.js
fidian: 00fcf10

spec/lib/formatter/generic-formatter.spec.js
```diff
@@ -0,0 +1,112 @@
+"use strict";
+
+describe("formatter/genericFormatter", () => {
+    var defaultTransform, genericFormatterAsync, resMock;
+
```
lnagle93: Is there a reason for having two line breaks here?

spec/lib/formatter/generic-formatter.spec.js
```diff
@@ -0,0 +1,112 @@
...
+
```
lnagle93: Is there a reason for having two line breaks here?

lnagle93:
Left one comment about an extra line break. Besides that, everything looks fine to me.

spec/lib/formatter/generic-formatter.spec.js
fidian: We talked about the whitespace rules in person.  The short answer is yes, functions should be offset by two lines from anything near them.  Comments are stuck to the code below them, so the whitespace moves up above the comment.

lib/formatter/error-json-formatter.js
AbsentSemicolon: Cool, sounds good.

AbsentSemicolon:
Reviewed and looks good.

PR #106, opened 2016-08-29
Signatures

fidian:
Please squash and merge this commit.  This commit requires the error-formatter branch to be merged first.

This is primarily signature validation.

* Middleware that processes a request and verifies that the basic elements for a signed request is present.
* Bulk of work done by version-specific signature module.

Also:

* Changed the mock error response logRef property to say "fakeLogRef" to preserve the capitalization of the name.
* New utility functions to pull out the duplicated code.
* OtDate objects are now immutable.  Previously, calling `.plus()` would alter the original date object, making it difficult to test for date windows or to report on a date after it was modified.
* New HMAC calculation routine in the hash library.

This does *NOT* add the signature validation to any routes.  That is going to be a separate pull request.  Unfortunately, that also means that this one isn't functionally tested and there still may be errors in the software.


fidian:
AGI-642

lib/middleware/validate-signature-middleware.js
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        return promise.resolve(result);
+    }
+
+
+    /**
+     * Parses the Authorized header and verifies the signature
+     *
+     * @param {Restify~req} req
```
lnagle93: There's an inconsistency between this line and 207. 207 seems a tad more explicit: 
`* @param {Restify~Request} req`

lib/middleware/validate-signature-middleware.js
```diff
@@ -0,0 +1,267 @@
...
+            title: "signature-information"
+        }
+    };
+
+    /**
+     * Middleware that calls authenticate and then either signals that
+     * the request is signed or else indicates the error and prevents
+     * future middleware from running.
+     *
+     * @param {boolean} [allowUnsigned=false]
```
lnagle93: I think this is the first time I've seen something like `[allowUnsigned=false]` in JSDoc. When/why should it be used?

lib/signature-ot1.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+
+    /**
+     * We only allow one method, algorithm and encoding.
+     *
+     * @param {validateSignatureMiddleware~signatureDescription} desc
+     * @return {Promise.<*>}
+     */
+    function checkSignatureDescriptionAsync(desc) {
```
lnagle93: On line 228, you used 'signatureDesc' as the param. For the sake of consistency, one should be changed.

lib/middleware/validate-signature-middleware.js
fidian: Agreed.  17a87df

lnagle93:
Besides my one comment on lib/signature-ot1, everything looks good to me.

lib/middleware/validate-signature-middleware.js
fidian: You use brackets when a parameter is optional.  If it is optional and defaults to a specific value then you can make that very explicit by documenting it.  Normally I don't care too much for things like `options`, but this one is special and I want to be very clear that if you don't pass `true` then unsigned requests are *not* allowed.

lib/signature-ot1.js
fidian: Agreed.  993732f

lib/formatter/binary-formatter.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+module.exports = (genericFormatter) => {
+    return genericFormatter.formatWithFallback((req, res, body, done) => {
+        if (!Buffer.isBuffer(body)) {
+            body = new Buffer(body.toString(), "binary");
```
AbsentSemicolon: Wondering if we should have a buffer library or something of that nature, since every time we do a buffer it's binary, or a config option saying to have it in binary so we don't always have to write it and if we decide to use some other form of buffer format we can change it easier.

lib/formatter/image-png-formatter.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+module.exports = (genericFormatter) => {
+    return genericFormatter.formatWithFallback((req, res, body, done) => {
+        if (!Buffer.isBuffer(body)) {
+            body = new Buffer(body.toString());
```
AbsentSemicolon: Do we not want to format into `"binary"` here?

lib/middleware/validate-signature-middleware.js
```diff
@@ -0,0 +1,267 @@
...
+     * @return {function} middleware
```
AbsentSemicolon: `function` should be capitalized.

fidian: Fixed in 3258c49

lib/formatter/image-png-formatter.js
fidian: UTF8 is the default.  You're right, I need to use "binary".

AbsentSemicolon:
Done with first review. Please see comments.

AbsentSemicolon:
Reviewed and looks good.

PR #107, opened 2016-08-29
Guards

fidian:
This adds functions that ensure that things like an accountId is passed to the managers.  If a required parameter is not passed, this throws an error and that will result in a failed Promise.

Also, for standardization, functions were renamed.

* delAsync -> deleteAsync
* destroyAsync -> deleteAsync

Finally, there are a few bugfixes and adjustments.

* The type checking in chain-middleware.js was fixed to compare against the string "undefined".
* Comments were added to ciphers-and-hashes.js indicating what hashes are known to be insecure.
* If a session is not validated it returns 403 Forbidden instead of 401 Unauthorized.
* The `routesMatch` option for Restify Router Magic needed to be specified because there is a shared bit of code that provides a "GET" response for a couple routes.
* `url` is changed to `uri` because that's closer to right.


fidian:
AGI-642
Must be merged after the signatures branch.

fidian:
Sorry, forgot to mention that I am not yet done with the branch.  I will update this when the branch is ready.

fidian:
Ready for review.

lib/access-code-manager.js
```diff
@@ -22,6 +22,37 @@ module.exports = (config, promise, random, storageServiceFactory) => {
     var codeLength, secretLength, storageService;
 
     /**
+     * Verifies that an account ID is not empty.
+     *
+     * @param {string} accountId
+     * @return {Promise.<*>}
+     */
+    function checkAccountIdAsync(accountId) {
```
lnagle93: This function is rewritten in account-manager, challenge-manager, and session-manager. Could it be moved into its own module to DRY things up?

lib/access-code-manager.js
```diff
@@ -22,6 +22,37 @@ module.exports = (config, promise, random, storageServiceFactory) => {
...
+
+
+    /**
+     * Verifies an account ID and a code are not empty.
+     *
+     * @param {string} accountId
+     * @param {string} code
+     * @return {Promise.<*>}
+     */
+    function checkBothAsync(accountId, code) {
```
lnagle93: This is rewritten in session-manager, with the only difference being the name of the parameter and the error message. Given that it's only written twice, it doesn't seem totally necessary to abstract it out of here and session-manager, unless it seems likely that it will be written a third time down the line. Cross that bridge when we get to it, perhaps?

lnagle93:
Besides the two comments above, everything looks good.

AbsentSemicolon:
Looks good to me, I agree with @lnagle93 and we should address the code areas being pretty much the same.

lib/access-code-manager.js
fidian: Agreed.  Moved guards to the storage service.  It isn't the manager's job to do validation.  Instead that should be done at the entry point (routes, command-line tools, etc) and exit points (storage engine in this case).  0c2459b

lib/access-code-manager.js
fidian: Moved to storage engine.  0c2459b

fidian:
I moved guards to the storage service.  This put them all into one place, which is good.  The managers should not be the ones that do validation.  Instead that should be done at entry/exit points.  The managers are the glue in the middle.

Routes validate already (JSON schemas and whatnot).  The output (storage engine) did not validate that keys were based on non-empty strings.  The last commit undid a lot of the work and instead moved it to the storage manager.

AbsentSemicolon:
Reviewed and looks good to me.

lnagle93:
Looks good. *thumbs up emoji*

PR #110, opened 2016-09-05
Adding tokenization and detokenization

fidian:
This commit is a bit bigger because I moved the managers to their own
folder.  We now have six of them so we should group them.


fidian:
AGI-650
@AbsentSemicolon @lnagle93

config.json
```diff
@@ -158,5 +158,19 @@
             "bucket": "opentoken-io-test-s3",
             "region": "us-east-1"
         }
+    },
+    "token": {
+        "idHash": {
+            "algorithm": "sha512",
+            "derivedLength": 32,
+            "encoding": "base64-uri",
+            "iterations": 4000,
```
AbsentSemicolon: Is 4000 enough, I thought we'd want to do 10000 at least. But I see it's using sha512 so maybe 4000 is enough.

lib/manager/token-manager.js
```diff
@@ -0,0 +1,98 @@
...
+        if (options.contentType) {
+            record.contentType = options.contentType;
+        }
+
+        if (options.public) {
+            record.public = true;
+        }
+
+        return random.idAsync(idLength).then((tokenId) => {
+            return storageService.putAsync([
```
AbsentSemicolon: Is it possible for tokens to overwrite another one? I know the hashes are astronomically large, but could it happen?

route/account/_account-id/token/index.js
```diff
@@ -0,0 +1,47 @@
...
+                readBodyBufferMiddleware(),
+                validateSignatureMiddleware(),
+                validateRequestQueryMiddleware("/account/token-create-request.json"),
+                (req, res, next) => {
+                    var publicFlag;
+
+                    publicFlag = false;
+
+                    // Query string values are always strings - compare as such
+                    if (req.query && req.query.public === "true") {
```
AbsentSemicolon: Are we sure we want to compare to the string? I know we validate by type as well, but maybe we should do `req.query.public.toLowercase() === "true"` as least, just to make sure somehow we don't get `TruE`.

spec/lib/middleware/validate-request-query-middleware.spec.js
```diff
@@ -0,0 +1,60 @@
...
+            middleware(req, res, (result) => {
+                expect(schemaMock.validate).toHaveBeenCalledWith({
+                    query: true
+                }, "schema");
+                expect(typeof result).toBe("undefined");
+                expect(res.send).not.toHaveBeenCalled();
+                done();
+            });
+        });
+        it("errors when the schema does not validate", (done) => {
```
AbsentSemicolon: Will these not work with the promises helper so you don't have to call `done`, or is because they are middleware we need to be more specific with them and call `done`?

example/signed-request.sh
```diff
@@ -1,30 +1,103 @@
...
+#/ Make a signed request to an OpenToken API.  This is a proof of concept
+#/ illustrating that signed requests can be made using only the Bash shell.
+#/
+#/ Requires the following environment variables:
+#/
+#/ * OPENTOKEN_API: URL typically used for self discovery.
+#/   eg. http://api.example.com/
+#/ * OPENTOKEN_CODE: The API code.
+#/ * OPENTOKEN_SECRET: The API secret.
+#/ * DEBUG: When set to a non-empty value, this turns on some debugging.
```
lnagle93: What does 'some' mean, in this case?

lib/middleware/validate-request-query-middleware.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+/**
+ * Parse an incoming request and validate that the query matches a schema.
+ *
+ * Use this as middleware in your route files.
+ */
```
lnagle93: Documentation was put above the `"use strict"` in lib/middleware/read-body-buffer-middleware.js. Do we have a convention regarding the placement of documentation vs `"use strict"` at the top of files?

route/account/_account-id/token/_token-id/index.js
```diff
@@ -0,0 +1,45 @@
+"use strict";
+
+module.exports = (server, path, options) => {
```
lnagle93: `options, path, and server` should be alphabetized.

route/account/_account-id/token/index.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+
+module.exports = (server, path, options) => {
```
lnagle93: `options, path, and server` should be alphabetized.

config.json
fidian: This is the hash for mapping the incoming token to a different identifier.

As far as PBKDF2 goes, there doesn't seem to be any solid standard on the number of iterations.  Different sources recommend different levels.

I'm updating all of the iterations to be larger values in case they are < 10,000.

lib/manager/token-manager.js
fidian: Yes.

There's approximately a 1 in 64^32 (6.2E57) chance of a token being the same as another token in the system.  Keep in mind that the tokens are also per-account.

Very, very unlikely.  doc/strength-and-numbers.md discusses this specific number.

route/account/_account-id/token/index.js
fidian: All incoming query values are strings.  I can't compare against a boolean or anything else.

The values of the string are laid out in the schema.  See schema/account/token-create-request.json for details, but the string is limited to "true" and "false".  It is optional and acts like "false" when not specified.

spec/lib/middleware/validate-request-query-middleware.spec.js
fidian: I copied an existing test and modified it without thinking.  To fix this I updated all three middleware test files that did it in the old way.

example/signed-request.sh
fidian: Updated comment:

    #/ * DEBUG: When set to a non-empty value, this turns on debugging output,
    #/   including line-by-line output and showing the entire curl command.

lib/middleware/validate-request-query-middleware.js
fidian: Yes.  File descriptions (this type of comment) come before "use strict".  I've updated all middleware files.

route/account/_account-id/token/_token-id/index.js
fidian: Not for route files.  They do not use dependency injection and thus must use the parameter order as specified in [Restify Router Magic](https://github.com/tests-always-included/restify-router-magic#route-files) - just take a peek at the Route Files and later in that section where it talks about factory functions.

route/account/_account-id/token/index.js
fidian: Same as above.

fidian:
I force pushed an update so the newest changes are incorporated.

AbsentSemicolon:
Reviewed and looks good.

route/account/_account-id/token/_token-id/index.js
lnagle93: Good to know, thanks.

lnagle93:
Looks good to me.

PR #111, opened 2016-09-06
Binary buffer module

lnagle93:


lib/base32.js
```diff
@@ -41,9 +41,7 @@ module.exports = (thirtyTwo) => {
         // When passing data to this library, one must guarantee that the
         // data is NOT a string, otherwise an encoding like UTF-8 will be
         // assumed.
-        if (!Buffer.isBuffer(data)) {
-            data = new Buffer(data, "binary");
-        }
+        binaryBuffer.toBuffer(data);
```
fidian: Don't you want `data = binaryBuffer.toBuffer(data);`?

lnagle93: Looks like I may have had some issue pushing, because that fix was already made locally. I'll look into it.

Issue #109, opened 2016-08-31
Make a binaryBuffer module


fidian:
As [noted](https://github.com/opentoken-io/opentoken/pull/106/files/993732fdbed8b6be33936be22b4d729adae41221#r77002057) in #106, there's a lot of code that looks similar to this:

```
// Scenario 1: change string/Buffer into Buffer
if (!Buffer.isBuffer(x)) {
    x = new Buffer(x.toString(), "binary");
}

// Scenario 2:  change string/Buffer into string
if (Buffer.isBuffer(x)) {
    x = x.toString("binary");
}
```

The idea is to scrub the codebase of these things and move them into `lib/binary-buffer.js` and inject `binaryBuffer` everywhere.  That way we don't forget to use `"binary"` in key areas (as was also noticed in #106).  Proposed methods:

```
binaryBuffer.toBuffer(stringOrBuffer)
binaryBuffer.toString(stringOrBuffer)
```

These can be augmented or changed as is appropriate, depending on the needs of the code.

fidian:
This was closed by #112.

PR #112, opened 2016-09-06
Binary buffer module

lnagle93:
This pull request addresses this issue: https://github.com/opentoken-io/opentoken/issues/109


lib/binary-buffer.js
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Converts a string into a buffer. Does not change the input if it is
+     * already a buffer.
+     *
+     * @param {(Buffer|string)} stringOrBuffer
+     * @return {(Buffer|string)}
+     */
+    function toBuffer(stringOrBuffer) {
+        if (!Buffer.isBuffer(stringOrBuffer)) {
+            stringOrBuffer = new Buffer(stringOrBuffer.toString(), "binary");
```
fidian: For a tiny speed boost (considering how often this is used, it might be good), I suggest changing `stringOrBuffer =` into `return`.

fidian: You're calling `.toString()` on a string?

lib/binary-buffer.js
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Converts a buffer into a string. Does not change the input if it is
+     * already a string.
+     *
+     * @param {(Buffer|string)} stringOrBuffer
+     * @return {(Buffer|string)}
+     */
+    function toString(stringOrBuffer) {
+        if (Buffer.isBuffer(stringOrBuffer)) {
+            stringOrBuffer = stringOrBuffer.toString("binary");
```
fidian: Same as above, swap assignment with `return`.

spec/lib/base64.spec.js
```diff
@@ -1,10 +1,11 @@
 "use strict";
 
 describe("base64", () => {
-    var base64;
+    var base64, binaryBuffer;
 
     beforeEach(() => {
-        base64 = require("../../lib/base64")();
+        binaryBuffer = require("../../lib/binary-buffer")();
```
fidian: I'd prefer a mock to be made instead so I can later use `expect(binaryBuffer.toBuffer).toHaveBeenCalledWith("a string")`.

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
```
fidian: Filename is wrong.  Remove `-module`.

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    var binaryBuffer, copy;
+
+    beforeEach(() => {
+        binaryBuffer = require("../../lib/binary-buffer")();
+    });
+    [
+        "some string",
+        "",
+        new Buffer("some buffer"),
+        123
```
fidian: This input is odd.  I don't mind it being there but it isn't a string and it isn't a buffer, so you're intentionally violating the input as described in the jsdoc for the method.

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    ].forEach((scenario) => {
+        describe(".toBuffer()", () => {
+            it("should convert a string to a buffer", () => {
```
fidian: "should convert" -> "converts"
Be authoritative.

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
...
+                scenario = binaryBuffer.toBuffer(scenario);
+                expect(scenario).toEqual(jasmine.any(Buffer));
+            });
+            it("should not alter the stringOrBuffer if it is already a buffer", () => {
```
fidian: "should not alter the stringOrBuffer if" -> "preserves the input when"

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
...
+                copy = scenario;
+                scenario = binaryBuffer.toBuffer(scenario);
+                expect(scenario).toEqual(jasmine.any(Buffer));
+                expect(scenario).toBe(copy);
```
fidian: This test is effectively identical to the above.  If you really want to test behavior changes between the two, break out the scenarios.

fidian: It is interesting that this even works.  When copy is a number, why does `expect(scenario).toBe(copy)` pass?

spec/lib/binary-buffer-module.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        describe(".toString()", () => {
+            it("should convert a buffer to a string", () => {
+                scenario = binaryBuffer.toString(scenario);
+                expect(scenario).toEqual(jasmine.any(String));
+            });
+            it("should not alter the stringOrBuffer if it is already a string", () => {
+                copy = scenario;
+                scenario = binaryBuffer.toString(scenario);
+                expect(scenario).toEqual(jasmine.any(String));
+                expect(scenario).toBe(copy);
```
fidian: Again, these don't feel like different aspects of each scenario because the testing of the input isn't related to the testing of the output.

fidian:
Reviewed.

fidian:
Reviewed

lib/binary-buffer.js
lnagle93: Eslint is barking about the first change. Should I add `/* eslint-disable consistent-return */`to the top of the file as well (`// eslint-disable-next-line consistent-return` was not working)?

`.toString()` has been removed.

spec/lib/binary-buffer-module.spec.js
lnagle93: Fixed.

spec/lib/binary-buffer-module.spec.js
lnagle93: Is there ever a case where you'd want an input that violates the jsdoc (besides if you have error handling/are testing error handling)?

spec/lib/binary-buffer-module.spec.js
lnagle93: Fixed

spec/lib/binary-buffer-module.spec.js
lnagle93: Changed.

lib/binary-buffer.js
fidian: It would be better to figure out why eslint is barking instead of disabling anything.  Do not disable anything.  If it still fails, send me the code that has the failure.

spec/lib/binary-buffer-module.spec.js
fidian: Tests should cover all situations that you're expecting and a few that even should not happen.  Tests are a way of defining your routine's behavior.  So, maybe at times throw odd stuff at a routine like you did here - that's why I didn't say it was a bad idea.  I just think that the tests are broken because they are passing when 123 is used.

lib/binary-buffer.js
lnagle93: Got it figured out.

lib/binary-buffer.js
lnagle93: Fixed.

spec/lib/binary-buffer-module.spec.js
lnagle93: Gotcha.

They were passing as a result of the extraneous `.toString()` you mentioned in another comment. 123 has now been removed.

spec/lib/base64.spec.js
AbsentSemicolon: I agree, we should have a mock for binary buffer.

spec/lib/encoding.spec.js
```diff
@@ -4,11 +4,12 @@ describe("encoding", () => {
     var encoding;
 
     beforeEach(() => {
-        var base32, base64, hex;
+        var base32, base64, binaryBuffer, hex;
 
-        base32 = require("../../lib/base32")(require("thirty-two"));
-        base64 = require("../../lib/base64")();
-        hex = require("../../lib/hex")();
+        binaryBuffer = require("../../lib/binary-buffer")();
```
AbsentSemicolon: These aren't in alphabetical order. `ba` comes before `bi`.

spec/lib/binary-buffer-module.spec.js
lnagle93: f75db76

spec/lib/binary-buffer-module.spec.js
lnagle93: f75db76

lib/binary-buffer.js
lnagle93: I did find one test in which the `.toString()` method was having some influence. After removing `.toString()` in binary-buffer, this test in spec/lib/formatter/binary-formatter  breaks:

    it("essentially breaks with objects", () => {
         return formatterAsync({}).then((result) => {
             expect(Buffer.isBuffer(result)).toBe(true);                                                                                                                                                                                                                                                               
             expect(result.toString("binary")).toBe("[object Object]");
         });
    });

I was able to fix it by doing this: 

    it("essentially breaks with objects", () => {
        return formatterAsync({}).then(jasmine.fail, (err) => {                                                                                                                                                                                                                                                       
            expect(err.toString()).toBe("TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.");
        }); 
    });

I just wanted to make sure that this would be an ideal solution, or if there would be a better alternative.

fidian: I'd just expect that it would fail.

    it("breaks with objects", () => {
        return formatterAsync({}).then(jasmine.fail);
    });

lib/binary-buffer.js
fidian: You say fixed but I don't see updated code.

spec/lib/binary-buffer-module.spec.js
fidian: I think you need to push that change.

spec/lib/binary-buffer-module.spec.js
fidian: Not pushed.

Also, I would expect tests to be formatted like this:

    [
        "some string",
        new Buffer("some buffer")
    ].forEach((input) => {
        describe(".toBuffer()", () => {
            it("makes a buffer", () => {
                expect(Buffer.isBuffer(binaryBuffer.toBuffer(input))).toBe(true);
            });
            it("has the right content", () => {
                expect(binaryBuffer.toBuffer(input).toString()).toEqual(input.toString());
            });
        });
        describe(".toString(), () => {
            // test returned type
            // test returned value when converted to a string
        });
    });

This is super close to what you are currently doing, but the second test is worded badly.  You say something like "it should not convert the type" instead of "the value is correct".

spec/lib/base64.spec.js
lnagle93: f48ae01

lib/binary-buffer.js
lnagle93: bc09d95

spec/lib/binary-buffer-module.spec.js
lnagle93: 9345ec2

fidian:
Approved!

PR #113, opened 2016-09-06
Various fixes to eliminate errors of multiple types

fidian:
It was possible to have `.trim()` operate on a non-string before, so I
introduced checks to ensure I have a value instead of `undefined.

Fixed the wording in the documentation.

Updated an example script to have the right permissions and to not read
from stdin any longer.


fidian:
@AbsentSemicolon @lnagle93 AGI-650 for this one too.

fidian:
Nevermind - it doesn't have enough coverage.

PR #114, opened 2016-09-08
Various fixes to eliminate errors of multiple types

fidian:
It was possible to have `.trim()` operate on a non-string before, so I
introduced checks to ensure I have a value instead of `undefined.

Fixed the wording in the documentation.

Updated an example script to have the right permissions and to not read
from stdin any longer.

Fixed where request.body may be a Buffer and it should be changed to a
binary string instead of a UTF-8 encoded string (UTF-8 is the default
when typecasting Buffers to strings).

Ignore Elastic Beanstalk CLI config files.

Added deployment instructions.

Updated config to work as the production config.

Added a specialized health check route.

Removed an unnecessary property from the config.

Prevent nginx from logging requests.

Using Mac-acceptable flags for the date command.

### DIFF
--- a/doc/deploy.md
+++ b/doc/deploy.md
@@ -1,0 +1,41 @@
+Deployment to Elastic Beanstalk
+===============================
+
+Tools required:
+
+* git
+* gpg
+* node and npm
+* zip
+
+This guide is a bit terse and assumes that your hard drive is encrypted.  We don't want encryption keys leaking out to the world because a laptop was stolen.
+
+1. Clone the repository somewhere new.
+
+        git clone https://github.com/opentoken-io/opentoken.git release
+
+2. Install modules.
+
+        cd release
+        npm install
+
+3. Decrypt the encryption key.
+
+        gpg --decrypt encryption.key.asc > encryption.key
+
+4. Make sure tests pass.
+
+        npm test
+
+5. If that fails, **stop here**.
+6. Create an archive.
+
+        zip -r9 ../opentoken-$(date --utc +"%Y-%m-%d-%H-%M-%S").zip * .[^.]* --exclude .git/\* --exclude coverage/\*
+
+7. Login to the [AWS Console](https://opentoken-io.signin.aws.amazon.com/console).
+8. Upload the archive to Elastic Beanstalk.
+9. Wipe the release folder or at least delete `encryption.key`.
+
+Isn't it sweet using AWS for deployment?
+
+This whole process can be automated as well, but that's a task for another day.
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -1,0 +1,57 @@
+Getting Started
+===============
+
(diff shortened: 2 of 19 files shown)


doc/deploy.md
```diff
@@ -0,0 +1,41 @@
...
+
+3. Decrypt the encryption key.
+
+        gpg --decrypt encryption.key.asc > encryption.key
+
+4. Make sure tests pass.
+
+        npm test
+
+5. If that fails, **stop here**.
```
AbsentSemicolon: Should we have a step to say what to do if the tests fail. Something like, "figure out why and fix if possible"?

doc/getting-started.md
```diff
@@ -0,0 +1,57 @@
...
+        --description="Your description for these keys" \
+        --mfa=THE_MFA_SEED --password=REALLY_GOOD_PASSWORD
+
+This will just spit out a set of keys.  They consist of a code and a secret.  The code can be shared with others but keep the secret private!  The secret is used for signing requests.
+
+
+Tokenization, Detokenization
+----------------------------
+
+Signing a request to OpenToken.io is easier than AWS but based on the same principles.  There's a great request to do this for you as well.  Let's tokenize the file `tokenize-me.txt`.
```
AbsentSemicolon: Should add a comma after AWS.

doc/deploy.md
fidian: I somewhat hope that the person doing the deploy does not need to have that spelled out for them.  Am I assuming too much?

lib/signature-ot1.js
```diff
@@ -59,7 +59,12 @@ module.exports = (accessCodeManager, errorResponse, hash, promise, util) => {
         }
 
         result.push("");
-        result.push(req.body);
+
+        if (req.body) {
+            result.push(req.body.toString("binary"));
```
AbsentSemicolon: If this is a buffer, should we use `bufferBinary.toString(req.body)`?

AbsentSemicolon:
Reviewed, raised a few points.

doc/getting-started.md
fidian: I've used grammar checkers and they are not saying the same as you.

lib/signature-ot1.js
fidian: Absolutely!  It was not there when I wrote this code but it certainly is there now.

doc/deploy.md
AbsentSemicolon: I don't think so, just wanted to know if you thought it should be there, or missed it.

doc/getting-started.md
AbsentSemicolon: Very well, guess it just read weird in my head.

fidian:
Addressed the points, fixed the code for one thing.

lnagle93:
Looks good to me.

AbsentSemicolon:
Reviewed and looks good.

PR #115, opened 2016-09-15
Jasmine test helpers

lnagle93:


package.json
```diff
@@ -37,6 +37,7 @@
         "eslint": "^2.10.2",
         "istanbul": "^0.4.2",
         "jasmine-node": "^1.14.5",
+        "jasmine-node-helpers": "https://github.com/tests-always-included/jasmine-node-helpers",
```
fidian: We should wait until the repo is pushed to npm and renamed.  After that, this line should simply be

    "jasmine-test-helpers": "^1.0.0",

Or whatever the current version is.

lnagle93: Sounds good. I'll make that change.

AbsentSemicolon:
Looks good to me once we get the package published and can include it.

fidian:
Approved.  Squashing and merging.

fidian:
The one test is failing because of an issue introduced into Node.  https://github.com/nodejs/node/issues/8571

PR #118, opened 2016-09-30
Renaming lib/dependencies.js to lib/container.js

fidian:
This returns a configured container, so the new name seems to fit better.

Includes fixes for ESLint.

PR #119, opened 2016-09-30
Allowing configuration to be overridden

fidian:
You can now specify a whole config file by changing the one in the repository or point at an override file that is merged with the default settings.  This allows for configurations to be mostly standard and tweaked to be correct, such as when using the same software for dev, qa, and production.


fidian:
Must be after #118.
DOT-26.
Do not merge until I squash the commits, which will need to be done after 118 is merged.

lib/util.js
```diff
@@ -30,8 +37,46 @@ module.exports = () => {
...
+        if (typeof override !== "object") {
+            return override;
+        }
+
+        if (typeof base !== "object") {
+            return clone(override);
+        }
+
+        result = clone(base);
+        Object.keys(override).forEach((propName) => {
```
lnagle93: I believe there needs to be a line break above this line.

lnagle93:
Besides my one comment, everything looks good here.

bin/server.js
```diff
@@ -1,28 +1,105 @@
...
-if (process.env.DEBUG) {
-    logger.info("Environment variable DEBUG was set. DEBUG is on!");
-    config.debug = true;
+
+/**
+ * Loads the configuration
+ *
+ * @param {Dizzy} container
+ * @param {Object} args
+ * @return {Object} config
```
AbsentSemicolon: Should this be a typedef? I'm good with object because it's pretty big, just thought I'd ask.

config.json
```diff
@@ -134,10 +134,10 @@
         "storagePrefix": "session/"
     },
     "server": {
-        "baseUrl": "https://api.opentoken.io",
+        "baseUrl": "http://localhost",
```
AbsentSemicolon: Do you need to put this back to api.opentoken.io for deployment, or do you upload a different config when deploying?

lib/util.js
fidian: Nope.  That's only for `for`, `while`, `do`, `if`, and those sort of constructs.  In here we don't add newlines around a method call.

bin/server.js
fidian: I'm hesitant to put the typedef because it isn't the complete config.  It's an unprocessed one that could be anything.

config.json
fidian: I upload a different config.  Specifically, I use the `override.json.gpg` when uploading.

bin/server.js
AbsentSemicolon: Cool, just thought I'd ask.

config.json
AbsentSemicolon: Ok sounds good.

AbsentSemicolon:
Reviewed and looks good.

fidian:
Rebased off current master.  If all is well, merge (not squash).

AbsentSemicolon:
All looks good... merging.

PR #121, opened 2016-10-03
This now sends emails

fidian:


fidian:
This is the big pull request you were fearing.
DOT-206.
No dependencies.
Do not squash.

debug/render-template.js
```diff
@@ -0,0 +1,74 @@
...
+ * Resolves a path and loads a file with require()
+ *
+ * @param {string} filePath
+ * @return {*} whatever is returned from require()
+ */
+function loadFile(filePath) {
+    var resultPath;
+
+    resultPath = path.resolve(process.cwd(), filePath);
+    console.error(`Loading file: ${resultPath}`);
```
lnagle93: Why is this `console.error` rather than log or info?

lib/email/ses.js
```diff
@@ -0,0 +1,58 @@
...
+
+
+    /**
+     * Sends an email using a template and additional data.
+     *
+     * @param {string} recipient
+     * @param {string} subject
+     * @param {string} text
+     * @param {string} html
+     * @param {*} data
```
lnagle93: Is this data param necessary?

debug/render-template.js
fidian: Because I want it to go to stderr so I can pipe the output of the program to a file if needed.

lib/email/ses.js
fidian: No.  Removed.  Great catch.

debug/render-template.js
```diff
@@ -0,0 +1,74 @@
...
+dataFile: JSON file to read for template processing.  Use "-" for no data.
+override.json: Optional path to an override config to load.
+
+Sample:
+
+    # Render the template for the HTML portion of a registration email.
+    ./render-template ../template/email registration-html \\
+        email-data-samples/registration.json
+
+Renders the template to stdout.  Status messages to to stderr.`);
```
fidian: Yes.  Fixed.

lnagle93:
Reviewed.

schema/config/email.json
```diff
@@ -0,0 +1,46 @@
...
+    "required": [
+        "engine",
+        "from"
+    ],
+    "oneOf": [
+        {
+            "properties": {
+                "engine": {
+                    "enum": [
+                        "ses"
```
AbsentSemicolon: Wondering if we want to put enums for engines, what if they want to use something else? tThen they need to modify the code just to add a different engine.

AbsentSemicolon:
Reviewed and have one question.

schema/config/email.json
fidian: Yes, that is true.  We don't have a system to allow them to specify their own backend engines at this time and there's really no demand for it, so I don't see this as a problem.

AbsentSemicolon:
Reviewed and looks good.

PR #122, opened 2016-10-03
Fixing the HTTP status to be 201 Created

fidian:
This is more fitting than a 204.  There's nothing in the spec that says
201 must have a representation of the resource.  Also, 204 is more
general and can apply to more situations than 201.


fidian:
DOT-206, no dependencies.

lnagle93:
Reviewed. Looks good.

AbsentSemicolon:
Reviewed and looks good.

PR #123, opened 2016-10-04
Enabling overrides for production

fidian:
1. Switching from docopt to neodoc because the first one is stale and
does not work.  It's true that neodoc has its own issues, but at least
the author is responsive and the parsing can be made to work.

2. Reformatted help message in `bin/server.js` for neodoc a bit and had
to change how the args were accessed because of how neodoc returns them.

3. Tested `bin/server.js` and found that `config` was saved far too late
in the process.  Moved earlier.

4. Split up the scripts that were executed with `npm` and placed them
into their own directory.  This now means that `bash` must be installed
to get these things to work, but grants us much more flexibility in how
programs are executed and also gives us a great place for comments.


fidian:
DOT-206
No dependencies.
Do not squash.
A deployment can happen once this is merged!

fidian:
Pushed a fix to update the tests that I neglected to update.

lnagle93:
Reviewed, everything looks good.

lnagle93:
Looks good.

scripts/start
```diff
@@ -0,0 +1,16 @@
...
+
+# Check for the presence of an override file
+if [[ -e "override.json" ]]; then
+    startCommand+=("--override=override.json")
+fi
+
+# Start the server
+"${startCommand[@]}"
+
+# Even if the server died because of an exception or because of SIGINT,
```
AbsentSemicolon: `SIGINT` Signals Intelligence, or something else?

AbsentSemicolon:
Reviewed and looks good.

PR #129, opened 2016-12-14
Implementing random request IDs

### LINKED ISSUE #96: Make error responses more informational
The following functions have been identified as having the potential to trigger an error.  When these errors happen, they should provide more information to the API user instead of just saying that the it was a bad request or internal server error.  The user will not have enough information at their fingertips and won't be able to divine the way to correct the request.

```
challenge-manager.js - validateAsync()
validation-manager.js - secureAsync() - twice
validate-request-middleware
```

fidian:
* Request IDs feed into error responses as the logref.  This removes the previous implementation of randomly assigned logref values.
* Fixed promise generation during tests so Node 6.6 and later do not emit warnings about unhandled and asynchronously handled Promise objects.
* Fixed problems where errors were generated and no useful information was returned from the API.
* Removed CORS support since the plugin was broken.

This closes #96.


fidian:
Log time to DOT-313

fidian:
Updated to address code review concerns.  Also added a test to increase the branch-related test coverage.

lib/middleware/request-logger-middleware.js
```diff
@@ -0,0 +1,45 @@
...
+        var d;
+
+        d = new Date();
+        logger.info(`${this.logId} ${d.toISOString()} ${message}`);
+    }
+
+
+    idLength = config.server.requestIdLength;
+
+    /**
```
quantumew: I think another newline above the JSDoc and one less newline above `idLength = config.server.requestIdLength;`

lib/storage/s3.js
```diff
@@ -9,22 +9,20 @@
...
  *
- * @typedef {Object} s3~configOptions
- * @property {string} bucket
- * @property {string} region
+ * @typedef {Object} storage~putOptions
+ * @property {string} [contentType]
+ * @property {OtDate} [expires]
  */
 
 /**
```
quantumew: Two newlines above these JSDoc blocks.

fidian:
Fixed everything except the spaces after a function.  The style guide says there should be 2 blank lines after functions.  Weirdly, that note is under "Comments".  https://tests-always-included.github.io/style-guide/general/#comments

fidian:
Merged!

PR #132, opened 2017-02-24
Need to reset the container

fidian:
This prevents normal behavior from creeping in during our tests. We want
weird behavior because we're repatching the container's
genericFormatter, but we only want that weird behavior while running one
of these tests.

Before, the genericFormatter was cached automatically by Dizzy and was
reused. I suspect that my recent changes to Dizzy during the enabling of
promises caused the caching technique to work better (in Dizzy's eyes)
and thus break these tests.

### DIFF
--- a/spec/helper/formatter-to-promise.helper.js
+++ b/spec/helper/formatter-to-promise.helper.js
@@ -1,5 +1,9 @@
 "use strict";
 
+var mockRequire;
+
+mockRequire = require("mock-require");
+
 /**
  * Load an OpenToken-style formatter and wrap it in a Promise to make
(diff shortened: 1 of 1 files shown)


spec/helper/formatter-to-promise.helper.js
```diff
@@ -22,7 +26,8 @@ jasmine.formatterToPromise = (formatterName, reqMock, resMock) => {
         resMock = require("../mock/response-mock")();
     }
 
-    container = require("../../lib/container");
+    // Get a fresh copy of the container
```
GoogilyBoogily: Need a newline above this.

PR #135, opened 2017-03-14
Switching to Jasmine 2 and function route tester

quantumew:


package.json
```diff
@@ -36,11 +36,13 @@
         "chalk": "^1.1.3",
         "codecov": "^1.0.1",
         "eslint": "^3.6.0",
+        "functional-route-tester": "https://github.com/quantumew/functional-route-tester",
```
quantumew: NOTE: Shimmed until this gets published so build passes.

spec/functional/account.spec.js
```diff
@@ -46,8 +46,8 @@ describe("account", () => {
             });
         }).then((response) => {
             expect(response.statusCode).toBe(200);
-            expect(response.headers["Content-Type"]).toBe("text/plain");
-            expect(response.body.toString("binary")).toBe("This is a test token");
+            expect(response.getHeader("Content-Type")).toBe("text/plain");
+            expect(response.body.toString()).toBe("This is a test token\n");
```
fidian: This looks wrong. The body of the request (line 30) does not have a newline but this response includes a newline. That seems to indicate OpenToken is adding a newline to the payload instead of keeping it verbatim, and that's incorrect behavior.

This same feedback applies to the addition of a newline on line 84.

spec/helper/functional-test.helper.js
```diff
@@ -348,14 +314,30 @@ class FunctionalTest {
                 }
             });
         }).then((response) => {
-            this.state.sessionId = response.body.sessionId;
+            var body;
+
```
fidian: You've added great error checking elsewhere. Would you add it here as well or does that break tests?

spec/helper/functional-test.helper.js
```diff
@@ -543,17 +407,26 @@ class FunctionalTest {
...
         }
 
-        queryString = options.url.split("?");
-        path = queryString.shift();
-        queryString = queryString.join("?");
+        path = options.url;
+        queryString = querystring.stringify(options.query);
+
+        if (options.parameters) {
+            Object.keys(options.parameters).forEach((key) => {
```
fidian: I believe the functional route tester handles this. I don't think it needs to be done in OpenToken.

quantumew: This is just to fix the path for signing the request. I am just changing something like this `/blah/blah/token{?public}` to `/blah/blah/token`. So there is some overlap but it was the easiest way to fix the path for signing the request.

spec/helper/functional-test.helper.js
quantumew: Good call. Done

spec/functional/account.spec.js
quantumew: I think you are right let me track that down.

quantumew: Looks like the newline is added in this one by lib/formatter/text-formatter.js on this line

```
return done(null, new Buffer(`${content}\n`, "binary"));
```

fidian: Aw, crud, that's totally right. Resolved.

spec/helper/functional-test.helper.js
fidian: Gotchya. Ok, resolved.

PR #139, opened 2018-06-21
Fixed a typo in the Schema for server.json

Neekoy:
schema/config/server.json had a typo which was preventing it from properly validating correct config override for HTTPS.


fidian:
Thanks for the update! I typically performed SSL termination at a gateway instead of with this service, which is why I did not notice this issue.

Issue #138, opened 2018-06-19
[ERR_ASSERTION]: async formatter for application/json requires callback to res.send()


Neekoy:
Hello,

I tried running the code locally, but unfortunately I'm getting the following error:

```
Unhandled rejection AssertionError [ERR_ASSERTION]: async formatter for application/json requires callback to res.send() (func) is required
    at foundFormatter (/home/neekoy/projects/opentoken/node_modules/restify/lib/response.js:449:20)
    at ServerResponse._findFormatter (/home/neekoy/projects/opentoken/node_modules/restify/lib/response.js:155:12)
    at ServerResponse.__send (/home/neekoy/projects/opentoken/node_modules/restify/lib/response.js:426:10)
    at ServerResponse.send (/home/neekoy/projects/opentoken/node_modules/restify/lib/response.js:278:17)
    at /home/neekoy/projects/opentoken/node_modules/restify/lib/server.js:978:29
```

I added some debug info to that function, and it would seem that most of the information is missing when the request is made:

```javascript
File: /home/neekoy/projects/opentoken/node_modules/restify/lib/response.js

300 /**
301  * internal implementation of send. convenience method that handles:
302  * writeHead(), write(), end().
303  * @private
304  * @private
305  * @param    {Number} [maybeCode] http status code
306  * @param    {Object | Buffer | Error} [maybeBody] the content to send
307  * @param    {Object} [maybeHeaders] any add'l headers to set
308  * @param    {Function} [maybeCallback] optional callback for async formatters
309  * @param    {Function} format when false, skip formatting
310  * @returns  {Object} returns the response object
311  */
312 Response.prototype.__send =
313 function __send(maybeCode, maybeBody, maybeHeaders, maybeCallback, format) {
314 
315     var self = this;
316     var isHead = (self.req.method === 'HEAD');
317     var log = self.log;
318     var code;
319     var body;
320     var headers;
321     var callback = {};
322 
323     console.log("Maybe code: " + maybeCode);
324     console.log("Maybe body: " + maybeBody);
325     console.log("Maybe headers: " + maybeHeaders);
326     console.log("Maybe callback: " + maybeCallback);
327     console.log("Maybe format: " + format);
```

The above console logs return the following:

```
Maybe code: AssertionError [ERR_ASSERTION]: async formatter for application/json requires callback to res.send() (func) is required
Maybe body: undefined
Maybe headers: undefined
Maybe callback: undefined
Maybe format: true
```

Some system info:
```
[...]

PR #140, opened 2018-08-07
Reverted opentoken-io/restify to master branch.

Neekoy:
The previously used 'proxywrap' branch had some breaking changes. This is related to #138


fidian:
Were you able to figure out the problem? I have not.

Because this is breaking the code, I'm inclined to merge this request. Is there a reason that the `node-restify` shouldn't be used instead of my outdated fork?

fidian:
I think that this isn't an issue any longer, but you are welcome to double check my work.

